PR #4, opened 2019-03-01
update VERSION to 1.1.1

phanak-sap:
This is a "bugfix" version only for including the corrected setup.py homepage.

This was catched sadly after was released pypi package for 1.1, with incorrect homepage URL. PyPI does not allow to reuse same distribution name (with good reasons, see [this GitHub issue )](https://github.com/pypa/packaging-problems/issues/74)


filak-sap:
Thank you!

PR #5, opened 2019-03-01
Update README.md - suggested installation from pypi

phanak-sap:


filak-sap:
Well done! Thank you!

PR #10, opened 2019-03-15
Service: added usefull property

kovnat:
Changes to be committed:
	modified:   .gitignore
	modified:   pyodata/v2/service.py

Added property names to the class Service. It returns list of names of
entity sets. Also this property added to the class EntityContainer.
This property should make access to this names easier.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *.pyc
 __pycache__
 *.swp
+.idea*
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -1134,6 +1134,12 @@ class EntityContainer:
             raise AttributeError(
                 'EntitySet {0} not defined in {1}.'.format(name, ','.join(list(self._entity_sets.keys()))))
 
+    @property
+    def names(self):
+        """Names of entity sets"""
+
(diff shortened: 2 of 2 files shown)


filak-sap:
Thank you for taking the time to prepare and open this Pull request! We really appreciate your efforts.

We understand your goals but we are afraid that the property `name` of the class `EntityContainer` might collide with an entity set from the loaded OData service. As you can see the class `EntityContianer` is a kind of thin wrapper around a dictionary of EntitySet names and provides the syntactic sugar that avoids the need to address a dictionary item in square brackets.

If you need to iterate over all EntitySets, may I know why you cannot do this:
```python
for es in service.schema.entity_sets:
     print(es.name)
```
or if you need the list of EntitySet names:
```python
entity_set_names = [es.name for es in service.schema.entity_sets]
```

If you don't insist on the property `names` of the class `EntityContainer`, you can still have the convenience property `entity_names` in the class `Service`. But I would like to ask you to rename it to `entity_set_names`.

kovnat:
Thank you for your answer!
Can you add this to the USAGE.md? I used my code because I didn't understand how to get this values, and I didn't find this in the Usage tips.

PR #28, opened 2019-06-11
Handle association set ends with same entity sets

lubosmj:
In the metadata document, it is allowed to declare same Entity Sets in
different ends of one Association Set. Therefore, we cannot utilize
dictionary to store such ends (values can be overwritten).

In this commit, there is added a new class AssociationSetEndRole which
represents one end within the Association Set. End roles are stored in
a list, in the same manner like end roles in the Association. Also,
each end now contains a reference to a corresponding entity set.

### DIFF
--- a/pyodata/v2/model.py
+++ b/pyodata/v2/model.py
@@ -898,17 +898,19 @@ class Schema:
                         'Association {} does not exist in namespace {}'
                         .format(assoc_set.association_type_name, assoc_set.association_type_namespace))
 
-                for key, value in list(assoc_set.end_roles.items()):
-                    # Check if entity set exists in current scheme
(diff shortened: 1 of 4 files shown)


pyodata/v2/model.py
```diff
@@ -1556,6 +1558,46 @@ def from_etree(association_node):
         return association
 
 
+class AssociationSetEndRole:
+    def __init__(self, role, entity_set_name):
+        self._role = role
+        self._entity_set_name = entity_set_name
+        self._entity_set = None
+
+    def __repr__(self):
```
filak-sap: Please, add `__str__` too or only that one.

PR #35, opened 2019-07-17
 Fix namespace parsing

mamiksik:


pyodata/v2/model.py
```diff
@@ -2072,13 +2083,24 @@ def parse(metadata_xml):
             mdf = io.BytesIO(metadata_xml)
         else:
             raise TypeError('Expected bytes or str type on metadata_xml, got : {0}'.format(type(metadata_xml)))
-        # the first child element has name 'Edmx'
-        edmx = etree.parse(mdf)
+
+        edmx: etree._ElementTree = etree.parse(mdf)
```
filak-sap: Woohoo, what does this syntax do?

pyodata/v2/model.py
```diff
@@ -2072,13 +2083,24 @@ def parse(metadata_xml):
...
+
+            if qname.localname == 'Schema' and qname.namespace in EDM_WHITELIST:
+                NAMESPACES['edm'] = qname.namespace
+
 
         # aliases - http://docs.oasis-open.org/odata/odata/v4.0/odata-v4.0-part3-csdl.html
         Edmx.update_global_variables_with_alias(Edmx.get_aliases(edmx))
 
         edm_schemas = edmx.xpath('/edmx:Edmx/edmx:DataServices/edm:Schema', namespaces=NAMESPACES)
+
```
filak-sap: Superfluous new line.

pyodata/v2/model.py
```diff
@@ -2072,13 +2083,24 @@ def parse(metadata_xml):
...
-        # the first child element has name 'Edmx'
-        edmx = etree.parse(mdf)
+
+        edmx: etree._ElementTree = etree.parse(mdf)
+
+        for node in edmx.iter():
+            qname = etree.QName(node.tag)
+
+            if qname.localname == 'Edmx' and qname.namespace in EDMX_WHITELIST:
+                NAMESPACES['edmx'] = qname.namespace
```
filak-sap: Please, log a warning if the namespace is not on the whitelist.

pyodata/v2/model.py
```diff
@@ -2072,13 +2083,24 @@ def parse(metadata_xml):
...
+
+            if qname.localname == 'Schema' and qname.namespace in EDM_WHITELIST:
+                NAMESPACES['edm'] = qname.namespace
```
filak-sap: Please, log a warning if the namespace is not on the whitelist.

pyodata/v2/model.py
mamiksik: It mainly aids the autocomplete as the real signature isn't inherently known as lxml is not directly written in python.

pyodata/v2/model.py
mamiksik: removed

pyodata/v2/model.py
mamiksik: done

pyodata/v2/model.py
mamiksik: done

pyodata/v2/model.py
jfilak: Can we get rid of it?

mamiksik: done

tests/test_model_v2.py
```diff
@@ -807,3 +808,74 @@ def test_edmx_association_set_end_by_entity_set():
...
+
+def test_namespace_whitelist(metadata_builder_factory):
+    """Test correct handling of whitelisted namespaces"""
+
+    builder = metadata_builder_factory()
+    builder.namespaces['edmx'] = 'http://docs.oasis-open.org/odata/ns/edmx'
+    builder.namespaces['edm'] = 'http://docs.oasis-open.org/odata/ns/edm'
+    builder.add_schema('', '')
+    xml = builder.serialize()
+    Edmx.parse(xml)
```
filak-sap: There must be some assert or anything that makes sure it went OK or not?

pyodata/v2/model.py
```diff
@@ -2064,21 +2076,56 @@ def __init__(self):
...
+        if namespaces is not None:
+            NAMESPACES.update(namespaces)
+
+        xml = etree.parse(mdf)
+        edmx = xml.getroot()
+
+        try:
+            dataservices = next((child for child in edmx if etree.QName(child.tag).localname == 'DataServices'))
+        except StopIteration:
+            raise PyODataParserError('Metadata document is missing the element DataServices')
```
filak-sap: Is this line covered in our tests?

pyodata/v2/model.py
```diff
@@ -2064,21 +2076,56 @@ def __init__(self):
...
-        # the first child element has name 'Edmx'
-        edmx = etree.parse(mdf)
+
+        NAMESPACES['edmx'] = None
+        NAMESPACES['edm'] = None
+        del NAMESPACES['edmx']
+        del NAMESPACES['edm']
+
+        if namespaces is not None:
+            NAMESPACES.update(namespaces)
```
filak-sap: Is this line covered in our tests?

pyodata/v2/model.py
mamiksik: Line 859

pyodata/v2/model.py
mamiksik: Line 859

pyodata/v2/model.py
mamiksik: fce test_missing_data_service (line 813)

tests/test_model_v2.py
mamiksik: The function finishes with no exception raised so, the test passes otherwise it would fail. How can I assert for no exception in pytest to make this explicit?

jfilak: Yep, but even an empty function would succeed, right? So we either need a coverage checker or we need to make sure we actually tested something.

pyodata/v2/model.py
jfilak: OMG, sorry.

tests/test_model_v2.py
mamiksik: Agree. Do you have a tip for how to do it? (I haven't work with pytest before.) I tried to google it but I came up with nothing.

pyodata/v2/model.py
mamiksik: No worries. Is it otherwise ok?

tests/test_model_v2.py
jfilak: Mock the latest possible method call and check it was called:
https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once

pyodata/v2/model.py
jfilak: OK

pyodata/v2/model.py
jfilak: THX

pyodata/v2/model.py
```diff
@@ -2159,9 +2160,3 @@ def update_global_variables_with_alias(aliases):
             namespace, suffix = direction_key.rsplit('.', 1)
             for alias in aliases[namespace]:
                 SAP_VALUE_HELPER_DIRECTIONS[alias + '.' + suffix] = SAP_VALUE_HELPER_DIRECTIONS[direction_key]
-
-
-def schema_from_xml(metadata_xml):
```
filak-sap: This is a facade method and I would like to ask you to add it back because the goal was to hide the implementation details like we use the class `Edmx`.

docs/usage/initialization.rst
```diff
@@ -22,6 +22,26 @@ Basic initialization which is going to work for everybody:
     northwind = pyodata.Client(SERVICE_URL, requests.Session())
 
 
+Set custom namespaces
+---------------------------------
```
filak-sap: Please add a blank like below `----` and cut the line `-----` to the length of the caption.

pyodata/client.py
```diff
@@ -44,7 +44,7 @@ def __new__(cls, url, connection, odata_version=ODATA_VERSION_2):
 
             # create model instance from received metadata
             logger.info('Creating OData Schema (version: %d)', odata_version)
-            schema = pyodata.v2.model.Edmx.parse(resp.content)
+            schema = pyodata.v2.model.Edmx.parse(resp.content, namespaces)
```
filak-sap: Good job! I think we should switch to the method `schema_from_xml` in another PR ;)

pyodata/v2/model.py
mamiksik: ok. reverted

filak-sap:
Good job! I pushed these:
- e297ebf test: Fix code to be PEP8 complaint
- bdaf350 Add support for whitelisted and custom XML namespaces

mamiksik:
> Good job! I pushed these:
> 
> * [e297ebf](https://github.com/SAP/python-pyodata/commit/e297ebf4d24eb12c3b0081a76f7adadede611299) test: Fix code to be PEP8 complaint
> * [bdaf350](https://github.com/SAP/python-pyodata/commit/bdaf350c9a108e01dd321f40b393929b9b4b701c) Add support for whitelisted and custom XML namespaces

Thx. So what about the documentation? Should I perhaps create another PR?

filak-sap:
The documentation is part of the commit bdaf350
https://pyodata.readthedocs.io/en/latest/usage/initialization.html#set-custom-namespaces

Unfortunately, it's taking a prominent place. I would rather move it to Advance section because we don't expect many devs in need to use own XML namespaces.

PR #36, opened 2019-07-19
Adds Codecov

mamiksik:


filak-sap:
I am not sure about the file .coverage. Shouldn't we make it ignored by git?

filak-sap:
Would you mind adding also the codecov badge to README? It should be this URL: https://codecov.io/gh/SAP/python-pyodata/branch/master/graph/badge.svg

README.md
```diff
@@ -1,5 +1,6 @@
 [![Build Status](https://travis-ci.org/SAP/python-pyodata.svg?branch=master)](https://travis-ci.org/SAP/python-pyodata)
 [![PyPI version](https://badge.fury.io/py/pyodata.svg)](https://badge.fury.io/py/pyodata)
+[![codecov](https://codecov.io/gh/SAP/python-pyodata/branch/master/graph/badge.svg)](https://codecov.io/gh/mamiksik/python-pyodata)
```
jfilak: Hmm, is the url "https://codecov.io/gh/mamiksik/python-pyodata" ok?

mamiksik: We want the URL for SAP repo not mine fork, don't we?

mamiksik: ahh I see my bad

mamiksik:
> I am not sure about the file .coverage. Shouldn't we make it ignored by git?

According to the tutorial you provided we should include it. I think it is for future PR, where codecov will compare .coverage on master against result for travis and thus will provide us with information whatever the coverage went up or down.

filak-sap:
>  According to the tutorial you provided we should include it. ...

Only it contains your local file paths which looks strange.

mamiksik:
> > According to the tutorial you provided we should include it. ...
> 
> Only it contains your local file paths which looks strange.

Agree, it looks strange. I went over a couple of public repos which are using codecov and none of them had this file tracked. So I removed it, despite the tutorial. I guess there is a mistake in it.

filak-sap:
Good job! I squashed your branch and pushed the commit 6a533fc4c6c83d4beda2d8db04290c56e815aa0f

PR #52, opened 2019-08-21
No return type functions

filak-sap:
OData V2 FunctionImports do no have to have ReturnType defined:
https://help.sap.com/viewer/u_collaboration_dev_help/dacad2174f654e62a75754d71fee9da2.html#loiodacad2174f654e62a75754d71fee9da2__FunctionImport

### DIFF
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -1177,16 +1177,48 @@ class FunctionContainer:
         def function_import_handler(fimport, response):
             """Get function call response from HTTP Response"""
 
-            if response.status_code >= 300:
-                raise HttpError('Function import call failed with the status code {0}'.format(response.status_code),
+            if 300 <= response.status_code < 400:
+                raise HttpError(f'Function Import {fimport.name} requires Redirection which is not supported',
+                                response)
+
+            if response.status_code == 401:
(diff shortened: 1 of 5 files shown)


pyodata/v2/service.py
```diff
@@ -1177,10 +1177,17 @@ def __getattr__(self, name):
         def function_import_handler(fimport, response):
             """Get function call response from HTTP Response"""
 
-            if response.status_code != requests.codes.ok:
-                raise HttpError('Function import call failed with status code {0}'.format(response.status_code),
+            if response.status_code >= 300:
```
mnezerka: Wouldn't it be better to handle `Redirects` interval [300, 399] separately and report that current transport layer (requests) doesn't support redirects?. Reporting result as `failed` could be misleading in some cases. And of course, leave current `failed` reporting for both client range [400-499] and server range >500.

filak-sap: Thanks for the comment! I forgot I want to change error handling to something less stupid.

filak-sap:
@mnezerka Can you please have a look at the last commit I have just pushed to this PR.

pyodata/v2/service.py
filak-sap: I added one more commit which is supposed to make error reporting less confusing.

PR #57, opened 2019-09-05
Add delete entity function

### LINKED ISSUE #54: Function to delete entity?
Is there a delete entity function?

mamiksik:
We never realized that delete entity function has been missing until
issue 54 was created. Hence this commit which adds that functionality to
our lib.

All added code is tested and 'How to use delete function' was also added
to the documentation.

resolves #54


filak-sap:
Good job! Thank you!

PR #60, opened 2019-09-27
Fix datetime handling while creating/updating entity

mamiksik:


jfilak:
I am sorry but I cannot find a reference supporting your thesis that JSON DateTime values are different that the URL DateTime values.

pyodata/v2/model.py
```diff
@@ -376,6 +379,13 @@ def to_literal(self, value):
 
         return super(EdmDateTimeTypTraits, self).to_literal(value.isoformat())
 
+    def to_json(self, value):
+        if isinstance(value, str):
+            return value
+
+        # Converts datetime into timestamp in milliseconds
+        return f'/Date({int((value - datetime.datetime(1970, 1, 1)).total_seconds()) * 1000})/'
```
jfilak: datetime.timestamp(value) * 1000

tests/test_service_v2.py
```diff
@@ -586,6 +586,15 @@ def test_update_entity(service):
     assert isinstance(request, pyodata.v2.service.EntityModifyRequest)
 
     request.set(Value=34)
+    # Tests if update entity is correctly call to_json method
```
mnezerka: Typo: .... correctly calls

tests/test_service_v2.py
```diff
@@ -586,6 +586,15 @@ def test_update_entity(service):
...
+    request.set(Date=datetime.datetime(2017, 12, 24, 19, 0))
+
+    assert request._values['Value'] == 34
+    assert request._values['Date'] == '/Date(1514142000000)/'
+
+    # Tests if if passed perforated datetime, to json does not update given value (for backward compatibility reasons)
```
mnezerka: Would be nice to rephrase this comment - e.g. explanation of what is `perforated datetime` would be good, same for `to json`, is it name of the method?

pyodata/v2/model.py
mamiksik:
``` python
import datetime

# From datetime to timestamp
date = datetime.datetime(2017, 12, 24, 19, 0)

r1 = int(date.timestamp() * 1000)
print(r1)
# outputs: 1514138400000
# That is in GTM 2017/12/24 18:00 or in GTM+1 2017/12/24 19:00

r2 = int((date - datetime.datetime(1970, 1, 1)).total_seconds()) * 1000
print(r2)
# outputs: 1514142000000
# That is in GTM 2017/12/24 19:00 or in GTM+1 2017/12/24 20:00


# I deducted we have to send the datetime in GTM format, because 1) There is another type called datetimeoffset which is
# for passing datetime with timezones 2) Also passing it in different timezone would yield strange result.
# See the example below

# Timestamp 1514142000000 represents 2017/12/24 19:00 GTM. Let's see what happens if we repeatably parse this value to
# datetime and back to timestamp
t = 1514142000000


def foo(timestap):
    value = datetime.datetime(1970, 1, 1) + datetime.timedelta(milliseconds=int(timestap))
    return value.timestamp() * 1000


for i in range(0, 10):
    t = foo(t)

print(t)
# outputs: 1514106000000
# After around 10 iteration the timestamps differ for about 36000000 milliseconds

# Let's reset the value and trie different parsing method
t = 1514142000000


def foo2(timestap):
    value = datetime.datetime(1970, 1, 1) + datetime.timedelta(milliseconds=int(timestap))
    return int((value - datetime.datetime(1970, 1, 1)).total_seconds()) * 1000


for i in range(0, 10):
    t = foo2(t)

print(t)
# outputs: 1514138400000
# After around 10 iteration the timestamps does not differ.
# Therefore I think this is correct method to parse datetime as parser should yield original value if fed it's output

```

filak-sap: I am sorry, why did you paste that code?

mamiksik:
> I am sorry but I cannot find a reference supporting your thesis that JSON DateTime values are different that the URL DateTime values.

What do you mean by "URL DateTime"? Maby my comment under your next comment answers your question.

pyodata/v2/model.py
mamiksik: You can replicate my results yourself as enclosed code is runnable.

filak-sap: Ah, so it is not accurate.

mamiksik: Yes, thus I suggest not to replace it with int(datetime.timestamp(value) * 1000). Do you agree?

filak-sap: well, if we want to stick to seconds only, we can `int(datetime.timestamp(value)) * 1000`

filak-sap:
>> What do you mean by "URL DateTime"? Maby my comment under your next comment answers your question.

to_literal is used for converting values to URL parameters - which has the format `datetime".."`

filak-sap:
https://www.odata.org/documentation/odata-version-2-0/overview/

```
Edm.DateTime
Represents date and time with values ranging from 12:00:00 midnight, January 1, 1753 A.D. through 11:59:59 P.M, December 9999 A.D.

datetime'yyyy-mm-ddThh:mm[:ss[.fffffff]]' NOTE: Spaces are not allowed between datetime and quoted portion. datetime is case-insensitive

Example 1: datetime'2000-12-12T12:00'
```

pyodata/v2/model.py
mamiksik: I am a bit confused. I still think that the problem is due to timestamp handling of the timezones and not the rounding of floats. From my understanding, the timestamp in python contains timezone information but the ODATA definition of DateTime does not include that. The user would have to rather use DateTimeOffset. (DateTime was deprecated in ODATA V4 and solely replaced by DateTimeOffset to avoid this kind of confusion)

mamiksik:
> https://www.odata.org/documentation/odata-version-2-0/overview/
> 
> ```
> Edm.DateTime
> Represents date and time with values ranging from 12:00:00 midnight, January 1, 1753 A.D. through 11:59:59 P.M, December 9999 A.D.
> 
> datetime'yyyy-mm-ddThh:mm[:ss[.fffffff]]' NOTE: Spaces are not allowed between datetime and quoted portion. datetime is case-insensitive
> 
> Example 1: datetime'2000-12-12T12:00'
> ```

Also see: https://www.odata.org/documentation/odata-version-2-0/json-format/

image

filak-sap:
Sure, Martin, I am happy you provided me with the reference you used to implement the changes. Can you add it to the commit message, please?

1) I didn't know that URL.
2) All my comments were supposed to lead you to update the documentation.

I am sorry but I don't know OData protocol completely - hence a claim you fixed something should be acompanied with a reference proving that:

a) the previous implementation was wrong
b) your new implementation is good

(no matter I talked to you in person - we do this for history and for clarity - a reviewer should double check we didn't forget anything - and I could be wrong, which happens quite often)

pyodata/v2/model.py
filak-sap: You are right: https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp

```
Note

There is no method to obtain the POSIX timestamp directly from a naive datetime instance representing UTC time. If your application uses this convention and your system timezone is not set to UTC, you can obtain the POSIX timestamp by supplying tzinfo=timezone.utc:

timestamp = dt.replace(tzinfo=timezone.utc).timestamp()

or by calculating the timestamp directly:

timestamp = (dt - datetime(1970, 1, 1)) / timedelta(seconds=1)
```

So, I think that your implementation suffers from the same bug:
1. the variable value can hold datetime of whatever timezone
2. datetime.datetime(1970, 1, 1) returns datetime in your current timezone

I propose to use this:
`int(value.replace(tzinfo=timezone.utc).timestamp() * 1000)`

Please, don't forget to add a comment with the reference to the code - I think this information belongs to code.

I adapted your code:

```python
#!/usr/bin/python3

import datetime

# From datetime to timestamp
date = datetime.datetime(2017, 12, 24, 19, 0)

r1 = int(date.timestamp() * 1000)
print('date.timestamp=', r1)
# outputs: 1514138400000
# That is in GTM 2017/12/24 18:00 or in GTM+1 2017/12/24 19:00

r2 = int((date - datetime.datetime(1970, 1, 1)).total_seconds()) * 1000
print('date - datetime(1970, 1, 1)=', r2)
# outputs: 1514142000000
# That is in GTM 2017/12/24 19:00 or in GTM+1 2017/12/24 20:00


# I deducted we have to send the datetime in GTM format, because 1) There is another type called datetimeoffset which is
# for passing datetime with timezones 2) Also passing it in different timezone would yield strange result.
# See the example below

# Timestamp 1514142000000 represents 2017/12/24 19:00 GTM. Let's see what happens if we repeatably parse this value to
# datetime and back to timestamp
t = 1514142000000


def foo(timestap):
    value = datetime.datetime(1970, 1, 1) + datetime.timedelta(milliseconds=int(timestap))
    return value.replace(tzinfo=datetime.timezone.utc).timestamp() * 1000


for i in range(0, 10):
    t = foo(t)

print('timestamp * 10=', t)
# outputs: 1514106000000
# After around 10 iteration the timestamps differ for about 36000000 milliseconds

# Let's reset the value and trie different parsing method
t = 1514142000000


def foo2(timestap):
    value = datetime.datetime(1970, 1, 1, tzinfo=datetime.timezone.utc) + datetime.timedelta(milliseconds=int(timestap))
    return int((value - datetime.datetime(1970, 1, 1, tzinfo=datetime.timezone(datetime.timedelta(hours=1)))).total_seconds()) * 1000


for i in range(0, 10):
    t = foo2(t)

print('subtraction * 10=', t)
# outputs: 1514138400000
# After around 10 iteration the timestamps does not differ.
# Therefore I think this is correct method to parse datetime as parser should yield original value if fed it's output
```
```
date.timestamp= 1514138400000
date - datetime(1970, 1, 1)= 1514142000000
timestamp * 10= 1514142000000.0
subtraction * 10= 1514178000000
```
BTW: In the first iteration, I didn't set tzinfo to the subraction base datetime and I got this error:

```Traceback (most recent call last):
  File "timestamp.py", line 50, in <module>
    t = foo2(t)
  File "timestamp.py", line 46, in foo2
    return int((value - datetime.datetime(1970, 1, 1)).total_seconds()) * 1000
TypeError: can't subtract offset-naive and offset-aware datetimes
```

mamiksik: I also noticed is it not the same case with timezones for to_literal?

jfilak: Please, if you want to fix something else, create a solo commit.

mamiksik:
I think we are ready to merge.

PR #63, opened 2019-10-07
fix typo in tests

sergeibershadsky:
fix typo in test


filak-sap:
Good catch. Thank you!

Issue #65, opened 2019-10-10
Delete Request in Batch Mode, need header


ChengQian1007:
Hello,

when we run delete requests in Batch Mode, it raises an exception:

...service.py", line 57, in encode_multipart
for hdr, hdr_val in req.get_headers().items():
AttributeError: 'NoneType' object has no attribute 'items'

So we need to define the get_header() method for EntityDeleteRequest(ODataHttpRequest), e.g. like this:

def get_headers(self):
  return {'Accept': 'application/json'}

I would like to pull a request and do the bugfix.

filak-sap:
@mamiksik PTAL

PR #66, opened 2019-10-10
[wip] Fix: Delete requests need headers to be included in Batch

ChengQian1007:
Delete requests need headers to be included in Batch

Ref to Issue: #65

### DIFF
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -496,6 +496,9 @@ class EntityDeleteRequest(ODataHttpRequest):
     def get_method(self):
         # pylint: disable=no-self-use
         return 'DELETE'
+    
+    def get_headers(self):
+        return {'Accept': 'application/json'}
 
 
 class EntityModifyRequest(ODataHttpRequest):


filak-sap:
@ChengQian1007 Thank you for taking the time to analyze the problem and post the patch. Do you think you can add/update a test case for this change?

filak-sap:
@ChengQian1007 Thank you for taking the time to analyze the problem and post the patch. Do you think you can add/update a test case for this change?

filak-sap:
@ChengQian1007  I am sorry but I cannot find any reference for your statement. Which OData service implementation do you use?

filak-sap:
D'oh, I've just noticed your issue #65. Well, I am not sure if every Batch request must send some Headers and if it's a good idea to add a random dummy headers just to work around a weak implementation of Batch request generator.

What about to either return and empty dict or better detect None in the function encode_multipart?

ChengQian1007:
@filak-sap I use CDS View generated OData Service.
Thanks for the advices, I will find out whether a normal Delete Request contains header.

phanak-sap:
Closing because this is forgotten PR.. and most probably wrong (does not take into account possibility of Batch requests for data in ATOM format)

PR #70, opened 2019-10-23
Enhance the pyodata executable to parse local metadata and specify policies

filak-sap:
Every time we get an error report with metadata, we have to create a custom code to get the real error.

This PR should allow us to use the helper pyodata and avoid the need to write code.

### DIFF
--- a/bin/pyodata
+++ b/bin/pyodata
@@ -4,9 +4,27 @@ import sys
 from argparse import ArgumentParser
 
 import pyodata
+from pyodata.v2.model import PolicyFatal, PolicyWarning, PolicyIgnore, ParserError, Config
+
 import requests
 
 from getpass import getpass
+
+
+ERROR_POLICIES={
+    'FATAL': PolicyFatal,
+    'WARNING': PolicyWarning,
+    'IGNORE': PolicyIgnore
+}
+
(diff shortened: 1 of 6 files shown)


bin/pyodata
```diff
@@ -95,16 +121,58 @@ def _main(argv):
...
-    except pyodata.exceptions.HttpError as err:
-        sys.stderr.write(str(err))
-        sys.stderr.write('\n')
-        sys.exit(1)
-
-    print('[Fetching $metadata ...]')
-    client = pyodata.Client(args.SERVICE_ROOT_URL, session)
+    # Oh, I am sorry for the double negation,
+    # but I cannot find a better construction.
+    if args.no_session_init == False:
```
mamiksik: maybe `if not args.no_session_init`

mamiksik: otherwise LGTM

filak-sap:
Pushed:
b399f8a bin: enable specifying error policies
547fc1e bin: add the parameter disabling HTTP bootstrapping
9fec3f2 bin: support reading metadata from local file
5a61f80 client: add the possibility to provide static metadata
22cdfb2 build: add make targets for reporting coverage

PR #89, opened 2020-04-07
bump requests to 2.23 and move to install_require

phanak-sap:
This PR addresses two issues:

1. Pinned usage of two year old version of Requests lib - update from 2.20 to 2.23

2. Move of this dependency to install_requires/requriments.txt from tests_require/dev-requirements.txt

Reason: while mainly used in tests, requests package is imported and used in service.py as well. e.g.   https://github.com/SAP/python-pyodata/blob/8937eca8a36aa49b5dc34e2b8de08a7671b69ccc/pyodata/v2/service.py#L840

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,4 @@
-requests==2.20.0
+requests==2.23.0
 pytest>=2.7.0
 responses>=0.8.1
 setuptools>=38.2.4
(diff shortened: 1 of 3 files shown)


filak-sap:
My idea was to not have a hard dependency on requests and the import in service.py is wrong.
Also it is no good idea to have 1 commit per 2 changes.

phanak-sap:
That's valid concern - I wanted to make a point with this PR, since it is just few lines.  I will change this PR to be handling only the problem no. 1.

"not have a hard dependency on requests and the import in service.py is wrong."  - exactly, since it is solvable by some local constant with status code ranges, I will for new just create an issue for point 2 to be solved in some another PR in future.

phanak-sap:
point no 1. not handled correctly in git, closing for new PR
issue #90 created for point no.2

PR #92, opened 2020-04-10
Replace python builtin exceptions with custom ones

mamiksik:
This commit is divided into these subtasks:
1. Categorizing all exceptions into groups[1][2]
2. Creating custom exceptions(all are annotated directly in file exceptions.py)
3. Replacing all raise EXCEPTION with newly created ones
4. Resolving dependent try-catch block, i. e. replace builtin exceptions in catch statements.

The 4th task is a bit tricky. I had to change the code a bit more to reflect things such as certain catch block being redundant(model line 1170, 1074-1078, 1155), multiple previously different exceptions being now caught as one(model lines 1191-1210)

Tests have been changed accordingly. In a few places, assert_logging_policy was replaced by caplog. It is a preferable practice and doing so made it easier to adjust the test.

Also, a bug causing the test silently passing when it should not have was found on the line 1245-1250.

[1] https://github.com/mamiksik/python-pyodata/wiki/Exceptions
[2] https://github.com/mamiksik/python-pyodata/wiki/Exceptions-rewrite

### DIFF
--- a/pyodata/exceptions.py
+++ b/pyodata/exceptions.py
@@ -12,6 +12,28 @@ class PyODataModelError(PyODataException):
     """Raised when model error occurs"""
 
 
+class PyODataRuntimeError(PyODataException):
+    """ Raised when an error accures after initial state is set.
+        e. g. when trying to set entity type for EntitySet for the second time
+    """
+
+
+class PyODataValueError(PyODataException):
+    """ Raised when an value is invalid
+        e. g. malformated input, value is out of range, entity has two properties with same name etc
+    """
+
+
+class PyODataKeyError(PyODataException):
+    """ Raised when nonexistent element is requested
+        e. g. property, entity set etc...
+    """
+
+
+class PyODataTypeError(PyODataException):
+    """ Raised when type error occurs"""
+
+
 class PyODataParserError(PyODataException):
     """Raised when parser error occurs"""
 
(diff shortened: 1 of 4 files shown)


pyodata/exceptions.py
```diff
@@ -12,6 +12,28 @@ class PyODataModelError(PyODataException):
...
+        e. g. property, entity set etc...
+    """
+
+
+class PyODataTypeError(PyODataException):
+    """ Raised when type error occurs"""
+
+
 class PyODataParserError(PyODataException):
     """Raised when parser error occurs"""
```
phanak-sap, line 39: ExpressionError and HttpError are now the only two exceptions inherited from PyODataException but without the PyOdata prefix. My opinion is that it should follow the rest for consistency, even that it is partially outside the scope of PR.

tests/test_model_v2.py
```diff
@@ -728,21 +727,18 @@ def test_annot_v_l_missing_e_t(mock_warning, xml_builder_factory):
...
     metadata.config.set_custom_error_policy({
         ParserError.ANNOTATION: PolicyWarning()
     })
 
-    metadata.build()
-    assert_logging_policy(mock_warning,
-                          'RuntimeError',
-                          'Target Type Dict of ValueHelper(Dict/Value) does not exist'
-                          )
+    with caplog.at_level(logging.WARNING):
```
phanak-sap, line 738: this is much nicer - usage of https://docs.pytest.org/en/latest/logging.html

filak-sap:
Can you create a table with 2 columns (old exception , new exception) where you where lines are errors, please.

filak-sap:
Split that huge commit to logical parts like:
- assert_logging_policy by caplog (with explanation)
- subtasks

I'm closing this PR. Please, create one with shorter commits.

PR #93, opened 2020-04-11
update travis.yml - check multiple lxml version

phanak-sap:
Since we provide pyodata as library and lxml is unpinned dependency, we should check at least major stable releases of lxml. Travis-CI has limit 200 concurrent jobs for open source, currently we use 3x8=24 combinations of python and lxml versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,21 @@ python:
   - "3.6"
   - "3.7"
   - "3.8"
+  
+env:
+  - LXML_VERSION=3.7.3
+  - LXML_VERSION=3.8.0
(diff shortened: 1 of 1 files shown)


phanak-sap:
6 out of 24 combinations fails consistently. Needs investigation.

phanak-sap:
Now we got 18 combinations of python version and lxml version passing. Build time is increased from 2:30 to 6 minutes which is imho still acceptable. If merged, I will later try to decrease the total time with stages so only test would be executed for all combinations and other post build actions (bandit, codecov) are run separately.

filak-sap:
Well done. Thank you!

PR #94, opened 2020-04-15
[DO NOT MERGE]: tmp: annotate exceptions

filak-sap:
@mamiksik  @phanak-sap Please, comment in the commit which serves the only purpose to mark exceptions we need to deal with.

See the file exceptions.txt and verify I didn't miss any case.

Then we should go through the exceptions again and classify them by the rules in exceptions.txt

### DIFF
--- a/pyodata/client.py
+++ b/pyodata/client.py
@@ -18,11 +18,13 @@ def _fetch_metadata(connection, url, logger):
                  resp.content)
 
     if resp.status_code != 200:
+        # ANNOT_EX: fatal runtime communication initialization error on service side
         raise HttpError(
             'Metadata request failed, status code: {}, body:\n{}'.format(resp.status_code, resp.content), resp)
 
     mime_type = resp.headers['content-type']
     if not any((typ in ['application/xml', 'text/xml'] for typ in mime_type.split(';'))):
+        # ANNOT_EX: fatal runtime communication initialization error on service side
         raise HttpError(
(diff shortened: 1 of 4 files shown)


mamiksik:
I think you did not miss any, however, looking at the code I think it would be good if a custom exception would be raised in method parse at line 159 if a value is None and in method property at line 1 222 when a property is not found. 

Also before you pushed this commit I compiled the list of changes for last(closed) PR as requested https://github.com/mamiksik/python-pyodata/wiki/Exception-change-list, but I think there is of little use for it now.

phanak-sap:
I think it is generally OK. I would like split the "service block" to distinguish somehow "metadata invalid" use case for usage pyodata for example only as metadata validator - therefore logically out of "service" block since there can be no networking at all, only local XML file parsing. 

I should be able to distinguish exceptions that came out during the phase "model building/metadata validation" from phase "runtime exception from model from valid/invalid metadata depeding on set policies". 

I should have clear subset of exceptions that can be thrown from metadata validation and if none of that is thrown, I consider the metadata valid.

Other possibility is to have really method for only metadata parsing and validation, and not be hidden under client initialization (in use case when I don't really want to use the generated model to access any data)

`pyodata.Client('NOT_VALID', None, metadata=local_metadata, config=custom_config)`

mamiksik:
@filak-sap @phanak-sap If there aren't any further objections should I implement the exception as described in this PR?

phanak-sap:
dropping 6y old stalled PR

PR #100, opened 2020-04-29
Nullable values

filak-sap:


CHANGELOG.md
```diff
@@ -11,9 +11,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - handle GET EntitySet payload without the member results - Jakub Filak
+- both Literal and JSON DateTimes has Timezone set to UTC - Jakub Filak
 
 ### Fixed
 - removed superfluous debug print when parsing FunctionImports from metadata - Jakub Filak
+- use correct type of deserialization of Literal (URL) structure values - Jakub Filak
+- null values are correctly handled and nullable configuration is honored - Jakub Filak
```
phanak-sap, line 19: small note: does that means that actually line 19 is the topic of this PR and line 14 is another feature in same PR? 
only provoking a bit :)

pyodata/v2/model.py
```diff
@@ -426,7 +426,7 @@ def from_literal(self, value):
                 except ValueError:
                     raise PyODataModelError('Cannot decode datetime from value {}.'.format(value))
 
-        return value
+        return value.replace(tzinfo=datetime.timezone.utc)
```
phanak-sap, line 429: I would recommend for future readability to define this as constant with docstring pointing to odata v2 spec, explaining why the conversion happens instead of just call in return statement. 

Rationale:
- datetimes and timezones are bug creators and pains anyway
- In odata v4 this is different and the DateTime is pruned for this exact timezone reason.

tests/test_model_v2_VariableDeclaration.py
```diff
@@ -0,0 +1,74 @@
+"""Tests of OData Model: class VariableDeclaration"""
+
+import pytest
+from pyodata.v2.model import VariableDeclaration, Types
+from pyodata.exceptions import PyODataException
+
+
```
phanak-sap, line 7: While it is good to have so many unit test, since the problem is "nullable is incorrectly parsed from metadata" I would consider adding to fixtures new XML with metadata for this case - small, few fields, just every nullable type present.

It may be a bit more readable from one test that is parsing real metadata where the problem is when it fails.

filak-sap: > nullable is incorrectly parsed from metadata

Metadata parsing was fixed in https://github.com/SAP/python-pyodata/commit/f9144b318b455534cbf2f14a0511c33bd3592c0a

> I would consider adding to fixtures new XML with metadata for this case - small, few fields, just every nullable type present.

Sure, I am happy to review your PR with such a test.

pyodata/v2/model.py
filak-sap: What conversion are you talking about? Default DateTime has tzinfo hodling None.
What OData V2 spec are you talking about? There are no mentions of Time zones in OData V2.

CHANGELOG.md
filak-sap: > small note: does that means that actually line 19 is the topic of this PR and line 14 is another feature in same PR?
only provoking a bit :)

It makes no sense to open a new PR for every small change especially when the little changes is blocking a bigger change. The unification of timezones is not a feature at all - it's rather a bug fix.

> But for real: Should't the conversion be configurable with UTC as sane default?

I am going to add the constant you mentioned somewhere. If you want to make it configurable, I am happy to review your Pull Request with that change.

pyodata/v2/model.py
filak-sap: The current version uses the function current_timezone(). If you want, you can think up a user configuration API which wouldn't be just a set of random functions.

CHANGELOG.md
filak-sap: Done, have a look at the first commit. I added the function current_timezone().

tests/test_model_v2_VariableDeclaration.py
phanak-sap: I am OK that we just agree on that - sure different PR.

pyodata/v2/model.py
phanak-sap: > What OData V2 spec are you talking about? There are no mentions of Time zones in OData V2.

Yes, it is quite hard to find true specification in the DateTime type in odata version 2.
[Odata.org](https://www.odata.org/documentation/odata-version-2-0/overview/)) pointed me towards 
[ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) or [W3C XML schema](https://www.w3.org/TR/xmlschema-2/#dateTime) and its lexical representations, which can specify timezone information in the suffix e.g. '2002-10-10T00:00:00+05:00'.

But seems that odata v2 expects just UTC (and even without daylight savings info, so one more reason for odata v4)

- Edm.DateTime '2000-12-12T12:00'. 
- Edm.DateTimeOffset '2002-10-10T17:00:00Z'.

But doesn't that means that similar conversion should not be only for DateTime but for DateOffset as well?

https://github.com/SAP/python-pyodata/blob/ce0b6cf6e43dbfeccdc5ea0b917c6780641fb160/pyodata/v2/model.py#L211

filak-sap: Yep, DateTimeOffset expects UTC. I didn't touch DateTimeOffset, did I?

PR #102, opened 2020-04-29
Correct parsing of 'Nullable' attributes

### LINKED ISSUE #98: nullable is incorrectly parsed from metadata
The following code is used to parse 'nullable' parameter from metadata:
`self._nullable = bool(nullable)` (VariableDeclaration line #606)

The problem is that 'nullable' parameter is passed as a string value ("true" or "false") or None. So `bool(nullable)` treats both "true" and "false" values as True. Moreover, the default or missing value (None) should be treated as nullable, but it's not.

homutov:
**Changes:**
- Consistent way of parsing 'Nullable' bool attributes
- Supports 'Nullable' attribute for properties (default: True)
- Supports 'Nullable' attribute for function import parameters (default: False)

Fixes #98

### DIFF
--- a/pyodata/v2/model.py
+++ b/pyodata/v2/model.py
@@ -1668,7 +1668,7 @@ class StructTypeProperty(VariableDeclaration):
         return StructTypeProperty(
             entity_type_property_node.get('Name'),
             Types.parse_type_name(entity_type_property_node.get('Type')),
-            entity_type_property_node.get('Nullable'),
+            attribute_get_bool(entity_type_property_node, 'Nullable', True),
             entity_type_property_node.get('MaxLength'),
             entity_type_property_node.get('Precision'),
             entity_type_property_node.get('Scale'),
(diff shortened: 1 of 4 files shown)


pyodata/v2/model.py
```diff
@@ -2364,7 +2364,7 @@ def from_etree(function_import_node, config: Config):
         for param in function_import_node.xpath('edm:Parameter', namespaces=config.namespaces):
             param_name = param.get('Name')
             param_type_info = Types.parse_type_name(param.get('Type'))
-            param_nullable = param.get('Nullable')
+            param_nullable = attribute_get_bool(param, "Nullable", False)
```
filak-sap, line 2367: Could you use single quotes, please?

homutov: Done

filak-sap:
Well done! Thank you very much indeed.

PR #110, opened 2020-06-26
Fixed filter not working for services that require URLEncoded filters

bartonip:
Some services need the filter query string to be URLEncoded for the query to work.

All SAP products support URL Encoded filter strings however not all OData v2 implementations support non URL Encoded strings.


filak-sap:
Thank you very much for your contribution to pyodata!

PR #111, opened 2020-06-26
A few improvements

bartonip:
- Added <, <=, >, >= operators for GetEntitySetFilter
- Made headers attribute on ODataHttpRequest actually affect the headers
- Allow PATCH or PUT to be specified for EntityModifyRequest

### DIFF
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -512,12 +512,17 @@ class EntityModifyRequest(ODataHttpRequest):
        Call execute() to send the update-request to the OData service
        and get the modified entity."""
 
-    def __init__(self, url, connection, handler, entity_set, entity_key):
+    def __init__(self, url, connection, handler, entity_set, entity_key, method="PATCH"):
         super(EntityModifyRequest, self).__init__(url, connection, handler)
         self._logger = logging.getLogger(LOGGER_NAME)
         self._entity_set = entity_set
         self._entity_type = entity_set.entity_type
         self._entity_key = entity_key
+
+        if method.upper() not in ["PATCH", "PUT"]:
+            raise ValueError("Method must be either PATCH or PUT")
+
+        self._method = method
 
         self._values = {}
 
@@ -531,7 +536,7 @@ class EntityModifyRequest(ODataHttpRequest):
 
     def get_method(self):
         # pylint: disable=no-self-use
-        return 'PATCH'
+        return self._method
 
     def get_body(self):
         # pylint: disable=no-self-use
@@ -1154,7 +1159,7 @@ class EntitySetProxy:
         return EntityCreateRequest(self._service.url, self._service.connection, create_entity_handler, self._entity_set,
                                    self.last_segment)
 
-    def update_entity(self, key=None, **kwargs):
+    def update_entity(self, key=None, method="PATCH", **kwargs):
         """Updates an existing entity in the given entity-set."""
 
         def update_entity_handler(response):
@@ -1172,7 +1177,7 @@ class EntitySetProxy:
         self._logger.info('Updating entity %s for key %s and args %s', self._entity_set.entity_type.name, key, kwargs)
 
         return EntityModifyRequest(self._service.url, self._service.connection, update_entity_handler, self._entity_set,
-                                   entity_key)
+                                   entity_key, method=method)
 
     def delete_entity(self, key: EntityKey = None, **kwargs):
         """Delete the entity"""
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -235,11 +235,11 @@ class EntityKey:
 class ODataHttpRequest:
     """Deferred HTTP Request"""
 
-    def __init__(self, url, connection, handler, headers={}):
+    def __init__(self, url, connection, handler, headers=None):
         self._connection = connection
         self._url = url
         self._handler = handler
-        self.headers = headers
+        self._headers = headers or dict()
         self._logger = logging.getLogger(LOGGER_NAME)
 
     @property
@@ -272,6 +272,12 @@ class ODataHttpRequest:
         # pylint: disable=no-self-use
         return None
 
+    def add_headers(self, value):
+        if not isinstance(value, dict):
+            raise TypeError("Headers must be of type 'dict' not {}".format(type(value)))
+
+        self._headers.update(value)
+
     def execute(self):
         """Fetches HTTP response and returns processed result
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,8 +6,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+- Specify PATCH or PUT method for EntityUpdateRequest - Barton Ip
+- <, <=, >, >= operators on GetEntitySetFilter - Barton Ip
+
 ### Fixed
 - URL encode $filter contents - Barton Ip
+- Headers attribute on ODataHttpRequest - Barton Ip
```
filak-sap, line 15: I am sorry but I cannot see any code for this last changelog entry.

filak-sap:
Well done! Thank you. I pushed these commits:

https://github.com/SAP/python-pyodata/commit/90016c21e4e50946a30f8daef8bd9fc68416afcf
https://github.com/SAP/python-pyodata/commit/a735037851a0379aca1548d3665f7c0fb164862c
https://github.com/SAP/python-pyodata/commit/5a1040a03c0c62379f82551d8eb3705ee2cdbbc5

pyodata/v2/service.py
```diff
@@ -235,11 +235,11 @@ def __repr__(self):
 class ODataHttpRequest:
     """Deferred HTTP Request"""
 
-    def __init__(self, url, connection, handler, headers=None):
+    def __init__(self, url, connection, handler, headers={}):
         self._connection = connection
         self._url = url
         self._handler = handler
-        self._headers = headers
+        self.headers = headers
```
bartonip, line 242: The changes happen in the following lines

pyodata/v2/service.py
```diff
@@ -284,7 +284,7 @@ def execute(self):
         # pylint: disable=assignment-from-none
         body = self.get_body()
 
-        headers = {} if self._headers is None else self._headers
+        headers = self.headers
```
bartonip, line 287: This one

pyodata/v2/service.py
```diff
@@ -351,7 +351,7 @@ def get_path(self):
         return self._entity_set_proxy.last_segment + self._entity_key.to_key_string()
 
     def get_headers(self):
-        return {'Accept': 'application/json'}
+        return {'Accept': 'application/json', **self.headers}
```
bartonip, line 354: This one

pyodata/v2/service.py
```diff
@@ -448,7 +448,7 @@ def get_body(self):
         return json.dumps(self._get_body())
 
     def get_headers(self):
-        return {'Accept': 'application/json', 'Content-Type': 'application/json', 'X-Requested-With': 'X'}
+        return {'Accept': 'application/json', 'Content-Type': 'application/json', 'X-Requested-With': 'X', **self.headers}
```
bartonip, line 451: This one

pyodata/v2/service.py
```diff
@@ -541,7 +546,7 @@ def get_body(self):
         return json.dumps(body)
 
     def get_headers(self):
-        return {'Accept': 'application/json', 'Content-Type': 'application/json'}
+        return {'Accept': 'application/json', 'Content-Type': 'application/json', **self.headers}
```
bartonip, line 549: This one

pyodata/v2/service.py
```diff
@@ -639,6 +644,7 @@ def get_headers(self):
 
         return {
             'Accept': 'application/json',
+            **self.headers
```
bartonip, line 647: This one

pyodata/v2/service.py
```diff
@@ -699,6 +705,7 @@ def get_method(self):
     def get_headers(self):
         return {
             'Accept': 'application/json',
+            **self.headers
```
bartonip, line 708: This one

pyodata/v2/service.py
```diff
@@ -1433,7 +1452,7 @@ def get_boundary(self):
 
     def get_headers(self):
         # pylint: disable=no-self-use
-        return {'Content-Type': 'multipart/mixed;boundary={}'.format(self.get_boundary())}
+        return {'Content-Type': 'multipart/mixed;boundary={}'.format(self.get_boundary()), **self.headers}
```
bartonip, line 1455: This one

filak-sap:
@bartonip  OMG, my apologies, I skipped that commit - I saw "M" at the commit summary, I my brain interpreted it as a merge commit.

pyodata/v2/service.py
```diff
@@ -235,11 +235,11 @@ def __repr__(self):
 class ODataHttpRequest:
     """Deferred HTTP Request"""
 
-    def __init__(self, url, connection, handler, headers=None):
+    def __init__(self, url, connection, handler, headers={}):
```
filak-sap, line 238: Please, don't use objects for default values. The default value is instantiated only once. In this case, once someone populates headers, all other consecutive calls of ODataHttpRequests() will receive the hedaers directory populate.

Or was that your intention?

pyodata/v2/service.py
filak-sap: Do you really need to make the member public? If so, I would prefer a property.

pyodata/v2/service.py
filak-sap: Here we should have `headers = dict(self.headers)` to make sure we don't polute someones dict.

pyodata/v2/service.py
```diff
@@ -284,7 +284,7 @@ def execute(self):
         # pylint: disable=assignment-from-none
         body = self.get_body()
 
-        headers = {} if self._headers is None else self._headers
+        headers = self.headers
 
         # pylint: disable=assignment-from-none
         extra_headers = self.get_headers()
```
filak-sap, line 290: If you keep the get_headers method as it is this patch, then we don't need the line 287 at all. But I would rather keep the line 287 and remove all occurences of slef.headers from the redefined get_headers. I don't think its a good idea to force descendant classes to use the headers class member to return the correct headers - this is up to the parent class.

pyodata/v2/service.py
bartonip: That was my intention for my use case, a bit hacky but works well for us without needing to set global params. 

However, in retrospect you are correct we should not be forcing the headers on all descendant classes as this is definitely not expected behaviour. 

I will revise this to use the dict type as the default value.

pyodata/v2/service.py
bartonip: Gotcha, this way we can also have a set_headers method that can append/update headers.

filak-sap:
I pushed a slightly modified and squashed version:
https://github.com/SAP/python-pyodata/commit/2d7cd1603212efbe07ecadaf9d887699e12c8110

I hope you don't mind I kept you as author even thought I changed the patch - I prefer clean history where every single commit make sense.

PR #112, opened 2020-06-26
service: add service wide configuration

filak-sap:
In the case where we need to update several entity sets,
we probably don't want to specify HTTP method for each of them.

This patch does nothing but enhances user convenience.


filak-sap:
@bartonip What do you think about the proposed level of configuration?

bartonip:
Love this, good job.

PR #114, opened 2020-06-29
Fixed Content-Length no longer accurate if rehydrated by FakeSocket

bartonip:
Due to the response_str encountering inadvertent mutation elsewhere in the code, the size slightly changes and with a large response, does not correctly get reported to HTTPResponse.

We manually specify the length of the response to the size of the BytesIO buffer and this gives HTTPResponse enough room to read to the end of the response.


filak-sap:
Any kind of test possible?

bartonip:
The `from_string` method fails because in its process of converting it to and from bytes and being passed around, some bytes are being added somewhere and causing `HTTPResponse.begin()` to seed the wrong `length` attribute. So to clean this up we just need to manually set `response.length` to how large the body is now after all its encoding and decoding.

filak-sap:
Thank you! The commit https://github.com/SAP/python-pyodata/commit/35b333d447b8613999ef4293c572fbc64645a3ba closes this PR.

PR #115, opened 2020-06-29
Response headers passed through to EntityProxy

bartonip:
A `headers` keyword argument has been added in `EntityProxy` to allow the passthrough of the response headers which includes the `ETag` of a particular entity.

This allows the ETag to be extracted from the entity and used in an If-Match header as required.

### DIFF
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -512,12 +512,17 @@ class EntityModifyRequest(ODataHttpRequest):
        Call execute() to send the update-request to the OData service
        and get the modified entity."""
 
-    def __init__(self, url, connection, handler, entity_set, entity_key):
+    def __init__(self, url, connection, handler, entity_set, entity_key, method="PATCH"):
         super(EntityModifyRequest, self).__init__(url, connection, handler)
         self._logger = logging.getLogger(LOGGER_NAME)
         self._entity_set = entity_set
         self._entity_type = entity_set.entity_type
         self._entity_key = entity_key
+
+        if method.upper() not in ["PATCH", "PUT"]:
+            raise ValueError("Method must be either PATCH or PUT")
+
+        self._method = method
 
         self._values = {}
 
@@ -531,7 +536,7 @@ class EntityModifyRequest(ODataHttpRequest):
 
     def get_method(self):
         # pylint: disable=no-self-use
-        return 'PATCH'
+        return self._method
 
     def get_body(self):
         # pylint: disable=no-self-use
@@ -1154,7 +1159,7 @@ class EntitySetProxy:
         return EntityCreateRequest(self._service.url, self._service.connection, create_entity_handler, self._entity_set,
                                    self.last_segment)
 
-    def update_entity(self, key=None, **kwargs):
+    def update_entity(self, key=None, method="PATCH", **kwargs):
         """Updates an existing entity in the given entity-set."""
 
         def update_entity_handler(response):
@@ -1172,7 +1177,7 @@ class EntitySetProxy:
         self._logger.info('Updating entity %s for key %s and args %s', self._entity_set.entity_type.name, key, kwargs)
 
         return EntityModifyRequest(self._service.url, self._service.connection, update_entity_handler, self._entity_set,
-                                   entity_key)
+                                   entity_key, method=method)
 
     def delete_entity(self, key: EntityKey = None, **kwargs):
         """Delete the entity"""
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -235,11 +235,11 @@ class EntityKey:
 class ODataHttpRequest:
     """Deferred HTTP Request"""
 
-    def __init__(self, url, connection, handler, headers={}):
+    def __init__(self, url, connection, handler, headers=None):
         self._connection = connection
         self._url = url
         self._handler = handler
-        self.headers = headers
+        self._headers = headers or dict()
         self._logger = logging.getLogger(LOGGER_NAME)
 
     @property
@@ -272,6 +272,12 @@ class ODataHttpRequest:
         # pylint: disable=no-self-use
         return None
 
+    def add_headers(self, value):
+        if not isinstance(value, dict):
+            raise TypeError("Headers must be of type 'dict' not {}".format(type(value)))
+
+        self._headers.update(value)
+
     def execute(self):
         """Fetches HTTP response and returns processed result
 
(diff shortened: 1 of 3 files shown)


pyodata/v2/service.py
```diff
@@ -750,6 +750,8 @@ def __init__(self, service, entity_set, entity_type, proprties=None, entity_key=
         self._cache = dict()
         self._entity_key = entity_key
 
+        self.headers = dict(headers) if headers else dict()
```
filak-sap, line 753: I am not happy from ad-hoc extensions of a class that is supposed to hold Entity.
What if make the method `execute()` smarter? E.g. accepting a parameter turning the return value into a tuple with `(the_entity, the_http_response)`?

bartonip: We could house the actual response inside a protected property called `_raw` or similar and then use `Entity.get_raw_response()` to get the response?

filak-sap: I am worried about storing the headers in instances because headers are valid only after the first HTTP request and they got invalidated with consecutive requests.

What about to add `execute(response_headers=my_headers_storage)` or `execute_with_response() -> (EntityProxy, Response)`?

bartonip: That is true, however since the `EntityProxy` attributes are immutable, we may as well just store the response that created the instance? 

You're right, it will be invalidated upon the next call, however the only case this causes and issue is if you're keeping an old instance of that entity around. If you're doing that, you're probably aware that it's not the newest version of the object anyway.

As long as the headers for the `EntityProxy` are updated per instance, I don't see a problem with having it on the `EntityProxy` itself?

What would be even cooler is if housed in an inner class is the original request and response, that would make debugging API calls even easier.

filak-sap: Can you please have a look at this?
https://github.com/filak-sap/python-pyodata-1/commit/df377449a5f175e3e6916aa5052ee6d909eaa744

I am reluctant to add headers to EntityProxy also to avoid growing memory footprint for all others who are not interested in HTTP headers.

bartonip: That looks good and I’m fine with that if you’re happy with it. 

One final poke at the bear: what if we added an `embed_response` kwarg to `ODataHttpRequest.execute` so we could call it like `request.execute(embed_response=True)` which would then include the response on the EntityProxy object?

filak-sap: How do you want to use the headers? Can you share a code snippet?

bartonip: I don’t have the repo in front of me but the general idea is like this. 

```
request = service.entity_sets.Customer.get_entity(Id=69)

customer = request.execute(embed_response=True)

request = service.entity_sets.Customer.update_entity(Id=69)

request.set(Name=“Tim”)
request.add_header({“If-Match”: customer.response.headers.get(“ETag”)})

request.execute()
```

filak-sap: Interesting. Is the ETag and If-Match something that should be supported by pyodata natively?

bartonip: [Yes it should.](https://help.sap.com/viewer/68bf513362174d54b58cddec28794093/7.51.6/en-US/bea0e62beca343fca78ded10f2937658.html)

However not all OData implementations require the If-Match header to be specified.

filak-sap: I think the class EntityProxy should receive a new method called update and the method should handle etag and if-match and key resolution. Because the code snippet is ugly - not your fault at all - just pyodata are not mature enough as we had no such a use case yet.

```
request = service.entity_sets.Customer.get_entity(Id=69)

customer = request.execute(embed_response=True)
customer.Name = 'Tim'
customer.update()
```

Well I am not sure about the above. Maybe, we can just make the current method update_entity smarter:

```

```
customer = service.entity_sets.Customer.get_entity(Id=69).execute()

customer.Name = 'Tim'

service.entity_sets.Customer.update_entity(customer).execute()
```

In both cases, the update entity request would contain only the field Name.

filak-sap: I propose:
- to enhance EntityProxy to remember ETag
- to enhance EntityProxy to remember changed values
- to enhance the method update_entity of the class EntitySetProxy to accept instance of EntityProxy and create request with If-Match header and modified fields

bartonip: Mate, absolutely love that proposal. Had a think about it overnight, I like your original proposal.

I think we should also reimplement how EntityProxy handles returning values and actually set the attributes on the instance rather than just returning them when `__getattr__` is called. This means that we can introspect the class using `dir` or `__dict__` and get a human readable listing of all available properties and values which is super valuable.

I think depending on the OData implementation, (I may be wrong but I think this is determined by PUT vs PATCH methods) we may need to send the whole object representation back. 

Other than that I think tracking ETag internally is a great idea, re-writing the EntityProxy class to have mutable attributes is a great idea and implementing the straight `.update()` function is also a great idea.

Let me know what you think about the above and we'll do the leg work for you.

mamiksik: Hi, Jakub contacted me on Monday and asked me if I could do the implementation. I would finish it if that's fine with you. (My apologies I did not mention it here right away)

I am not sure about the benefits of setting the attributes directly on the instance. Firstly, because you would still need to implement the `__getattr__`  to handle uncached properties. Secondly, it's hard to debug and work with classes as it is not obvious which properties are set programmatically and which are not.

bartonip: Isn’t the last part just a basic ORM design pattern? You keep a cache of what was programatically assigned and then you compare the values on your class to your cache? Thats how both SQLAlchemy and Django
implement this.

filak-sap: @bartonip How do you picture it should work? I don't want to force users to define mapping classes. SQLAlchemy and Django uses the mapping classes to create schema but we don't need any such a feature as our schema is provided by $metadata and we cannot do anything about it (I got inspired by python suds).

If you want to get list of available properties, then read EntityType assigned to the entity. We can add the method __dir__ to the class EntityProxy to make it more convenient for users.

> Why not just ingerit your whole class from dict?

Because the class is not dict. If you iherit from dict, you are dict :)

mamiksik: @bartonip The ETag PR is done. Please have a look at #116. I would be happy to hear your comments.

filak-sap:
Fixed by https://github.com/SAP/python-pyodata/pull/116

filak-sap:
Entities now have the property etag. Well, I think I made a mistake because it can collide with etag entity property.

PR #116, opened 2020-07-07
Add etag property to EntityProxy

mamiksik:


bartonip:
Yep that looks great!

pyodata/v2/service.py
```diff
@@ -684,14 +684,23 @@ class EntityProxy:
...
         self._logger = logging.getLogger(LOGGER_NAME)
         self._service = service
         self._entity_set = entity_set
         self._entity_type = entity_type
         self._key_props = entity_type.key_proprties
         self._cache = dict()
         self._entity_key = entity_key
+        self._etag = etag
+
+        if isinstance(proprties, dict):
```
filak-sap, line 697: You need to document this if statement. Why do you need it? It must be either in the commit message or a comment in the code.

mamiksik: The if was there because of that the properties is optional argument, but I now see that is not entirely the case. properties in some cases can be a list of dict! Good Catch!

filak-sap: In the case the variable proprties is a list, why don't you handle it? What's is supposed to happen? BTW all the  code below expects the argument proprties holds a dict.

mamiksik: But how would you handle it when the list represents multiple entities? Where the ETag would be located? On the first entity? On all of them?

> What's is supposed to happen? BTW all the code below expects the argument proprties holds a dict.

Yes, that it's why it is inside `if isinstance(proprties, dict):`. I don't understand what you are trying to say by it?

filak-sap: Why do we have list at first place and how we handle that list anyways?

filak-sap: Why do we create an entity proxy but pass a list of properties of multiple entities while one would expect just 1 entity?

mamiksik: I think the test_create_entity_nested_list test this behaviour specifically.

filak-sap: Hey Martin! You added that if statement, so you should know exactly why you did so. The test you mentioned have a list in a property value which represents a nested entity available via a navigation property. I would say that if we pass a list somewhere, then we have a bug. Please, do your job and exactly describe the case where we pass a list.

filak-sap: OK. I had a look into it and the list value is no relevant and is caused by an invalid test - the multiplicity of the navigation property is 1:1, so the service should not respond with a list. You can remove that if statement. We have to add more checks to make sure we detect such an invalid response.

mamiksik: Should I update the tests then too?

filak-sap: Definitely, you must. Please, in a separate commit which should be place before the current commit.

bartonip: How is this going? Would love to help push this out to release soon as the part of our app that uses the pyodata module is ready to be deployed!

filak-sap: The if statement has been removed as there was a bug in a test case which is fixed by:
https://github.com/SAP/python-pyodata/pull/116/commits/079ae2020c85929d732f0daa59f6e21bb3a1f78d

pyodata/v2/service.py
```diff
@@ -1337,7 +1350,12 @@ def get_entity_handler(response):
 
             entity = response.json()['d']
 
-            return EntityProxy(self._service, self._entity_set, self._entity_set.entity_type, entity)
+            try:
+                etag = response.headers.get('ETag', None)
+            except AttributeError:
```
filak-sap, line 1355: For the record, please explain this AttributeError catch statement or remove it.

pyodata/v2/service.py
```diff
@@ -1392,7 +1410,12 @@ def create_entity_handler(response):
 
             entity_props = response.json()['d']
 
-            return EntityProxy(self._service, self._entity_set, self._entity_set.entity_type, entity_props)
+            try:
+                etag = response.headers.get('ETag', None)
+            except AttributeError:
```
filak-sap, line 1415: For the record, please explain this AttributeError catch statement or remove it.

pyodata/v2/service.py
filak-sap: @bartonip This is a new blocker. @mamiksik added the except AttributeError statement but has no reliable explanation why it's needed and it makes no sense to me.

pyodata/v2/service.py
bartonip: This can just be removed because .get will never raise an AttributeError??

filak-sap:
@mamiksik  Are you going to remove the catch AttibuteError or are you done?

filak-sap:
Pushed:
- 3c8d03f service: correct batch response API for headers
- 1868980 tests: fix navigation prop multiplicity test
- 5642bb7 service: add etag property to EntityProxy

bartonip:
Release? 😬😬😬

jfilak:
Yep, 1.6.0. Take care1

Issue #120, opened 2020-08-03
Get a count of entities with inlinecount


s7oev:
Hello, I couldn't find any mention of "inlinecount" in the code.

The description of inlinecount can be seen in the [URI conventions of OData 2](https://www.odata.org/documentation/odata-version-2-0/uri-conventions/), section 4.9.

Is this feature not supported in pyodata?

filak-sap:
Hi @s7oev , than you for taking the time to report this issue. It's not supported yet but feel free to add it.

s7oev:
@filak-sap thanks for your quick reply!

I'd love to try and contribute, although I can't promise I'll be able to deliver a working functionality. 

Having said that, I actually managed to implement a basic inlinecount functionality already.

My only problem is how to effectively return this to the user. I.e., it would have to happen in the get_entities_handler() method, I believe.

```
 def get_entities_handler(response):
            """Gets entity set from HTTP Response"""

            if response.status_code != HTTP_CODE_OK:
                raise HttpError('HTTP GET for Entity Set {0} failed with status code {1}'
                                .format(self._name, response.status_code), response)

            content = response.json()

            if isinstance(content, int):
                return content

            entities = content['d']
            if isinstance(entities, dict):
                entities = entities['results']

            self._logger.info('Fetched %d entities', len(entities))

            result = []
            for props in entities:
                entity = EntityProxy(self._service, self._entity_set, self._entity_set.entity_type, props)
                result.append(entity)

            return result
```

But currently this method returns only a list of the entities. My very initial solution where the goal was only to see if I get the inlinecount successfully if it is requested was to add a new entry to the result as follows:

```
[...]
            entities = content['d']
            if isinstance(entities, dict):
                if '__count' in entities:
                    inlinecount = entities['__count']
[...]
            if inlinecount is not None:
                result.append({'Inlinecount': inlinecount})
```

Of course this is a very bad solution, as it mixes a list of entities with the inlinecount. I was thinking about making the result a dictionary if the inlinecount is requested, with keys e.g. result['inlinecount'] and result['entities'] but this is also a bad solution I feel because the function will return different types in different cases.

I'll continue thinking about the problem tomorrow, but was wondering if you (or anyone else) has some input with regards to this.

filak-sap:
[...]

PR #121, opened 2020-08-04
inlinecount support and ProgramError

s7oev:
Add support for inlinecount and a ProgramError exception as discussed in #120

### DIFF
--- a/pyodata/exceptions.py
+++ b/pyodata/exceptions.py
@@ -35,3 +35,7 @@ class HttpError(PyODataException):
         super(HttpError, self).__init__(message)
 
         self.response = response
+
+
+class ProgramError(PyODataException):
+    """Raised when an error in the program logic occurs"""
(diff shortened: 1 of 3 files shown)


jfilak:
@s7oev  any particular reason why closed this PR?

s7oev:
@jfilak Yes, sorry, was just about to comment. I noticed two issues with my tests:

1. I had the third property 'City' for the entity Employee (instead of 'LastName'). This was only a semantical issue, but still for consistency with other tests, I wanted to fix this;
2. The test for the total_count exception was poorly formulated. I had actually requested the inline count, but the test was passing (i.e., the exception saying it was not requested was raised) only because the mock GET response wasn't returning the count (while an actual GET request would). 

Both are quick-fixes but I'd like to check everything again now, just to make sure everything's 100% good before opening a new pull request.

PR #127, opened 2020-10-01
add new property is_valid for Schema class

phanak-sap:
Consider this an idea thrown away. If OK, there should definitely be added more tests for various invalid metadata files parsed and checked for expected result, but at the moment is IMHO enough just modifying existing test cases. 

Rationale for new property:

Policies Fatal, Warning a Ignore are modifying the behavior in a way that following is not possible to do (create simple validator)

1)	Go trough entire metadata for having all problems logged (exceptions thrown) - Fatal stops on first error
2)	Some kind of information that there are real problem in metadata - Warning will go to the end, but mix warning messages (unsupported annotation etc.) with real metadata parsing problem (various exceptions). Parsing logger is possible but ugly.
3)	Policies as they are so far are actually useful for many usages of Pyodata. But IMHO such new property for metadata state (valid/not_parsed_correctly) is useful on is own.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - Add inlinecount support - Stoyko Stoev
 - Add a ProgramError exception - Stoyko Stoev
+- Add is_valid schema property - Petr Hanak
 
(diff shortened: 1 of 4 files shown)


phanak-sap:
Build is failing due #130

PR #129, opened 2020-10-09
Update travis.yml - add pypy compiler test env

phanak-sap:
https://www.pypy.org/


phanak-sap:
Build is failing due #130

phanak-sap:
Build are now working on Travis, valid test failures - in all LXML + python version seems to be same

`======================== 4 failed, 228 passed in 30.91s ========================`

filak-sap:
So you want to merge and expect that people know that pypy3 can fail?

phanak-sap:
> So you want to merge and expect that people know that pypy3 can fail?

I do not know how can I be more specific than what already is stated in the PR. Nothing in this sentence is true. It is entirely the opposite.

1. this PR is marked as Draft. I do not want to merge it at the moment.
2. to be possible to merge, I obviously need to fix those 4 tests, ( it seems they have same root cause). I do not want to introduce already failing build.
3. even if this PR will have 100 percent PASS rate on entire build matrix, the pypy3 passing for every other future PR still should be optional. This is noted by previous comment _"Pypy3...should be marked as allow_failures in build matrix, even when the test will be fixed. Standard Python must pass, pypy should not block the PRs.",_  This means, in more words, that even after this is merged, `.travis.yaml `needs to have this build job marked by its directive `allow_failures ` - meaning in effect, that a) all python versions for future PRs must pass b) passing on pypi will be still optional, until stated otherwise. Any failures detected on CI will be just issues on me, until proven that the pypy is worth the effort. Then, maybe, in some future, pypy environment will be marked as mandatory to pass on every PR - if the maintenance effort will be lower than the benefits from speed gain. Note: even that is stated that pypy is on average quicker than cpython, there should be for sure some performance testing before such enforcement.

phanak-sap:
We use Github Actions only now, so closing with follow-up in #237

Issue #130, opened 2020-10-09
Builds are failing due pylint upgrade


phanak-sap:
Changelog for minor pylint release:
http://pylint.pycqa.org/en/latest/whatsnew/2.6.html

Flake8 changes between 3.8.3 and 3.8.4 are only bugfixes
https://flake8.pycqa.org/en/latest/release-notes/3.8.4.html#bugs-fixed

Latest build from master which is passing:

 3 months ago
flake8-3.8.3
pylint-2.5.3
Your code has been rated at 10.00/10

Builds which are now failing, even that no code change 
pylint-2.6.0
flake8-3.8.4
Your code has been rated at 9.71/10
Makefile:33: recipe for target 'lint' failed
make: *** [lint] Error 12
The command "make check" exited with 2.

 
**Open question - should we:**
a) pin dev-requirements.txt dependencies?
b) run also master branch builds periodically and not only PRs?

phanak-sap:
Builds are passing now. New rules are disabled and they should be enabled with respective code refactoring.

PR #132, opened 2020-10-12
Update .pylintrc - disable new check from 2.6.0

phanak-sap:
This is a quick fix for issue #130. Rules should be enabled with respective code refactoring.


phanak-sap:
Details about new rules introduced in pylint 2.6.0

https://pycodequ.al/docs/pylint-messages/r1725-super-with-arguments.html
https://pycodequ.al/docs/pylint-messages/w0707-raise-missing-from.html

http://pylint.pycqa.org/en/latest/whatsnew/2.6.html

Issue #138, opened 2020-11-30
Is there a way to specify sap-client in SCTS_CLOUD_API_ODATA_SRV?


ksaito1125:
Initializing pyodata.Client using the code below will result in an authentication error.
I want to connect to a 510 client. Can I specify sap-client = 510 to connect to in pyodata.Client?

```
>>> import requests, urllib3
>>> from pyodata.exceptions import PyODataException, HttpError
>>> import pyodata
>>> urllib3.disable_warnings()
>>> SERVICE_URL = 'https://' + server + '/sap/opu/odata/sap/SCTS_CLOUD_API_ODATA_SRV'
>>> session = requests.Session()
>>> session.verify = False
>>> session.auth = (user, passwd)
>>> try:
...     client = pyodata.Client(SERVICE_URL, session)
... except HttpError as e:
...     e.response.status_code
...     e.response.headers['set-cookie']
... 
401
'sap-usercontext=sap-client=951; path=/'
>>> 
```

ksaito1125:
I tried running it with curl.

```
$ curl -u $USER -k -I https://$SERVER/sap/opu/odata/sap/SCTS_CLOUD_API_ODATA_SRV/$metadata
HTTP/2 401 
set-cookie: sap-usercontext=sap-client=951; path=/
content-type: text/html; charset=utf-8
content-length: 9321
sap-system: SD1
www-authenticate: Basic realm="SAP NetWeaver Application Server [SD1/951]"
sap-server: true
sap-perf-fesrec: 10227.000000
```

This is the same as what is being done below.
https://github.com/SAP/python-pyodata/blob/1.7.0/pyodata/client.py#L14

I can connect to the correct client by specifying sap-client.

```
$ curl -u $USER -k -I https://$SERVER/sap/opu/odata/sap/SCTS_CLOUD_API_ODATA_SRV/$metadata?sap-client=510
HTTP/2 200 
set-cookie: sap-usercontext=sap-client=510; path=/
set-cookie: SAP_SESSIONID_SD1_510=*******; path=/; secure; HttpOnly
content-type: text/html
content-length: 0
sap-processing-info: ODataBEP=,crp=,RAL=,st=,MedCacheHub=,codeployed=,softstate=
sap-server: true
sap-perf-fesrec: 41539.000000
```

Is there a way to specify sap-client in pyodata.Client?

phanak-sap:
Hi @ksaito1125.
[...]

PR #139, opened 2020-12-01
Add sap_client parameter

ksaito1125:
Enabled to set sap-client from environment variable.
If not set, the operation is the same as before.

To set sap-client to 500, do as follows.

```
export PYODATA_SAP_CLIENT=500
```

### DIFF
--- a/pyodata/client.py
+++ b/pyodata/client.py
@@ -1,5 +1,6 @@
 """OData Client Implementation"""
 
+import os
 import logging
 import warnings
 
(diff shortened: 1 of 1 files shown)


ksaito1125:
Please link with Issue 138.

phanak-sap:
I am leaving this on hold, discussion in the #138.

phanak-sap:
alternative solution discussed in the issue #138 works OK

PR #141, opened 2020-12-17
doc: update initialization with sap-client example

ksaito1125:
Added documentation for issue #138


ksaito1125:
@phanak-sap
I tried to modify the documentation.
Thank you very much.

phanak-sap:
Thank you. It is perhaps somehow strange to combine Northwind and sap-client, but for the purpose of the code example it is good enough.

PR #144, opened 2021-02-20
chore: %-formatting to f-string

phanak-sap:
Because we have minimal Python version listed as 3.6, we can safely move to PEP 498 -- Literal String Interpolation.

f-strings are less verbose, more powerful (it is expression, e.g. can call functions) and quicker than old school string formatters. 

Converted using https://github.com/ikamensh/flynt

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -65,7 +65,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=locally-disabled, super-with-arguments, raise-missing-from, logging-fstring-interpolation
+disable=locally-disabled, super-with-arguments, raise-missing-from
 
 
 [REPORTS]
--- a/.pylintrc
+++ b/.pylintrc
@@ -65,7 +65,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=locally-disabled, super-with-arguments, raise-missing-from
+disable=locally-disabled, super-with-arguments, raise-missing-from, logging-fstring-interpolation
 
 
 [REPORTS]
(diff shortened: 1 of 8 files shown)


filak-sap:
Hi Peter, thank you very much for the time to create this fabulous PR. Unfortunately I have a bad news for you - the linting rule should not be disabled but must be respected and log messages must not be formatted with f-strings. Yeah, neither format is correct - I did it wrong. However, I cannot approve this PR because of the second commit for disabling of the pylint rule.

phanak-sap:
Hi @filak-sap - so you prefer mixing of string formatting styles? It is easy to revert it back, but the "rule" is just a warning, perf effect of switching to f-string is negligible and the ball is actually on the logging library. Also, we have already some rules disabled anyway.. so I don't quite understand why this additional disabled rule should be a problem. 

My view is that having all string in the same way is consisted, cleaner and therefore more readable. I would understand more if the problem would be stated that currently the linter job does not fail if PR would have old-style formatting .)

jfilak:
https://github.com/PyCQA/pylint/issues/1788#issuecomment-379180025

jfilak:
All is good, switch to f-string but don't turn off that warning and don't use f-strings in log mesages.

phanak-sap:
> All is good, switch to f-string but don't turn off that warning and don't use f-strings in log mesages.

We are now running in circles... the long comment of mine was explanation why I think it would be better to change even the three instances of f-strings in log messages (it is that many) for style consistency. I don't see any explanation why this rule should not be turned off (three others are already). I have no problem with doing that.. just want to know why I should.

Could you please elaborate why this commit is a problem for pyodata? It is not that "pylint must be respected" per se - we already ignore some rules, since it is not worth to adhere - https://github.com/SAP/python-pyodata/blob/e78a34c78aee1e26e6411d99983539e4e72c1ce8/.pylintrc#L68

We additionally have many pylint checks dispabled in the code itself, that it may be worth to revisit the .pylintrc entirely.

phanak-sap:
OK, done. Expected this to be squashed.

PR #154, opened 2021-05-14
Observer for http requests

mnezerka:
This change allows caller to observe technical details (e.g. headers, body, etc.) of http requests that are sent in the background of odata calls.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -20,7 +20,7 @@
 # -- Project information -----------------------------------------------------
 
 project = 'PyOData'
-copyright = '2019 SAP SE or an SAP affiliate company'
+copyright = '2021 SAP SE or an SAP affiliate company'
 author = 'SAP'
 
 # The short X.Y version
--- a/docs/usage/advanced.rst
+++ b/docs/usage/advanced.rst
@@ -95,3 +95,42 @@ it is enough to set log level to the desired value.
     logging.basicConfig()
     root_logger = logging.getLogger()
     root_logger.setLevel(logging.DEBUG)
+
+Observing HTTP traffic
+----------------------
+
+There are cases where you need access to the transport protocol (http). For
+example: you need to read value of specific http header. Pyodata provides
+simple mechanism to observe all http requests and access low level properties
+from underlying network library (e.g. **python requests**).
+
+You can use basic predefined observer class
+``pyodata.utils.RequestObserverLastCall`` to catch last response headers:
+
+.. code-block:: python
+
+    from pyodata.utils import RequestObserverLastCall
+
+    last = RequestObserverLastCall()
+    northwind.entity_sets.Employees.get_entity(1).execute(last)
+    print(last.response.headers)
+
+You can also write your own observer to cover more specific cases. This is an example of
+custom observer which stores status code of the last response.
+
+.. code-block:: python
+
+    from pyodata.utils import RequestObserver
+
+    class CatchStatusCode(RequestObserver):
+
+        def __init__(self):
+            self.status_code = None
+
+        def http_response(self, response, request):
+            self.status_code = response.status_code
+
+    last_status = RequestObserverLastCall()
+
+    northwind.entity_sets.Employees.get_entity(1).execute(last_status)
+    print(last_status.status_code)
--- a/pyodata/v2/model.py
+++ b/pyodata/v2/model.py
@@ -1018,6 +1018,8 @@ class Schema:
             except KeyError:
                 pass
 
+        raise KeyError('Association does not exist')
+
     @property
     def associations(self):
         return list(itertools.chain(*(decl.list_associations() for decl in list(self._decls.values()))))
@@ -1047,6 +1049,8 @@ class Schema:
                 return decl.association_sets[set_name]
             except KeyError:
                 pass
+
+        raise KeyError('Association set does not exist')
 
     @property
     def association_sets(self):
--- a/pyodata/utils/__init__.py
+++ b/pyodata/utils/__init__.py
@@ -1,0 +1,5 @@
+"""Utilities for Python OData client"""
+
+from .request_observer import RequestObserver, RequestObserverLastCall
+
+__all__ = ["RequestObserver", "RequestObserverLastCall"]
--- a/pyodata/utils/request_observer.py
+++ b/pyodata/utils/request_observer.py
@@ -1,0 +1,34 @@
+"""
+Interface for request observer, which allows to catch odata request processing details
+
+Author: Michal Nezerka <michal.nezerka@gmail.com>
+Date:   2021-05-14
+"""
+
+from abc import ABC, abstractmethod
+
+
+class RequestObserver(ABC):
+    """
+    The RequestObserver interface declares methods for observing odata request processing.
+    """
+
+    @abstractmethod
+    def http_response(self, response, request) -> None:
+        """
+        Get http response together with related http request object.
+        """
+
+
+class RequestObserverLastCall(RequestObserver):
+    """
+    The implementation of RequestObserver that stored request and response of the last call
+    """
+
+    def __init__(self):
+        self.response = None
+        self.request = None
+
+    def http_response(self, response, request):
+        self.response = response
+        self.request = request
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -22,6 +22,7 @@ except ImportError:
     from urllib2 import quote
 
 from pyodata.exceptions import HttpError, PyODataException, ExpressionError, ProgramError
+from pyodata.utils import RequestObserver
 from . import model
 
 LOGGER_NAME = 'pyodata.service'
@@ -296,7 +297,7 @@ class ODataHttpRequest:
 
         self._headers.update(value)
 
-    def execute(self):
+    def execute(self, observer: RequestObserver = None):
         """Fetches HTTP response and returns processed result
 
            Sends the query-request to the OData service, returning a client-side Enumerable for
@@ -319,6 +320,9 @@ class ODataHttpRequest:
         params = "&".join("%s=%s" % (k, v) for k, v in self.get_query_params().items())
         response = self._connection.request(
             self.get_method(), url, headers=headers, params=params, data=body)
+
+        if observer:
+            observer.http_response(response, response.request)
 
         self._logger.debug('Received response')
         self._logger.debug('  url: %s', response.url)
--- a/tests/test_service_v2.py
+++ b/tests/test_service_v2.py
@@ -6,6 +6,7 @@ import requests
 import pytest
 from unittest.mock import patch
 
+from pyodata.utils import RequestObserverLastCall
 import pyodata.v2.model
 import pyodata.v2.service
 from pyodata.exceptions import PyODataException, HttpError, ExpressionError, ProgramError
@@ -2372,3 +2373,33 @@ def test_custom_with_get_entity(service):
 
     entity = service.entity_sets.MasterEntities.get_entity('12345').custom("foo", "bar").execute()
     assert entity.Key == '12345'
+
+
+@responses.activate
+def test_request_observer(service):
+    """Test use of request observer"""
+
+    responses.add(
+        responses.GET,
+        f"{service.url}/MasterEntities('12345')",
+        headers={'h1-key': 'h1-val', 'h2-key': 'h2-val'},
+        json={'d': {'Key': '12345'}},
+        status=200)
+
+    last = RequestObserverLastCall()
+    entity = service.entity_sets.MasterEntities.get_entity('12345').execute(last)
+
+    assert last.request is not None
+
+    assert len(last.request.headers) > 0
+    assert 'Accept' in last.request.headers
+
+    assert last.response is not None
+    assert last.response.status_code == 200
+    assert len(last.response.headers) == 3
+    assert 'Content-type' in last.response.headers
+    assert 'h1-key' in last.response.headers
+    assert 'h2-key' in last.response.headers
+    assert  last.response.headers['Content-type'] == 'application/json'
+    assert  last.response.headers['h1-key'] == 'h1-val'
+    assert  last.response.headers['h2-key'] == 'h2-val'


phanak-sap:
@mnezerka PLS sign CLA licence for your user.

docs/conf.py
```diff
@@ -20,7 +20,7 @@
 # -- Project information -----------------------------------------------------
 
 project = 'PyOData'
-copyright = '2019 SAP SE or an SAP affiliate company'
+copyright = '2021 SAP SE or an SAP affiliate company'
```
phanak-sap, line 23: This could be nice, separate PR, immediately merged.

pyodata/v2/model.py
```diff
@@ -1048,6 +1050,8 @@ def association_set(self, set_name, namespace=None):
             except KeyError:
                 pass
 
+        raise KeyError('Association set does not exist')
+
```
phanak-sap, line 1054: This could be nice, separate PR, immediately merged, unrelated to the new observer class.

docs/usage/advanced.rst
```diff
@@ -95,3 +95,42 @@ it is enough to set log level to the desired value.
...
     root_logger = logging.getLogger()
     root_logger.setLevel(logging.DEBUG)
+
+Observing HTTP traffic
+----------------------
+
+There are cases where you need access to the transport protocol (http). For
+example: you need to read value of specific http header. Pyodata provides
+simple mechanism to observe all http requests and access low lever properties
+directly from underlying engine (**python requests**).
```
phanak-sap, line 105: This is not true. Pyodata does not have underlying engine (**python requests**) - it is networking library agnostic, dependency to requests is not direct in requirements.txt, by readme https://github.com/SAP/python-pyodata#usage it only expect Requests library Session-like object. Pyodata can work with other http networking libraries.

phanak-sap:
I strongly think that this PR - apart from lines marked for another PR - should be closed and not merged.

Rationale - unnecessary code and workaround without PR exists.

Point 1 - It bring unnecessary code, which may be even tied to requests library only, while pyodata does not strictly use [
requests library](https://pypi.org/project/requests/) only. Imagine aiohttp with internal event loop and the RequestObserverLastCall - may work, may not (not sure) but we would need to start write networking library compatibility tests. Anyway, new code to maintain forever.

Point 2 - for the problems at hand - get to the response HTTP code, headers etc at runtime, existing mechanism of Requests Hooks can be used, see [requests documentation](https://2.python-requests.org/en/master/user/advanced/#event-hooks)

Sample code:

```
import requests
import pyodata

def print_url(r, *args, **kwargs):
    print(str(r.status_code) + ' - ' + r.url)
    print(r.headers)
    print('-------')

SERVICE_URL = 'http://services.odata.org/V2/Northwind/Northwind.svc/'

s = requests.Session()
s.hooks['response'].append(print_url)

northwind = pyodata.Client(SERVICE_URL, s)

employee1 = northwind.entity_sets.Employees.get_entity(1).execute()
print(employee1.FirstName)
employee2 = northwind.entity_sets.Employees.get_entity(2).execute()
print(employee2.FirstName)
```

output:

```
(v) c:\github\pyodata-hook>python hooks.py
301 - http://services.odata.org/V2/Northwind/Northwind.svc/$metadata
{'Content-Type': 'text/html; charset=UTF-8', 'Location': 'https://services.odata.org/V2/Northwind/Northwind.svc/$metadata', 'Date': 'Tue, 15 Jun 2021 09:49:25 GMT', 'Content-Length': '0'}
-------
200 - https://services.odata.org/V2/Northwind/Northwind.svc/$metadata
{'Cache-Control': 'private', 'Content-Length': '39035', 'Content-Type': 'application/xml;charset=utf-8', 'Expires': 'Tue, 15 Jun 2021 09:50:26 GMT', 'Vary': '*', 'Server': 'Microsoft-IIS/10.0', 'DataServiceVersion': '1.0;', 'X-AspNet-Version': '4.0.30319', 'X-Powered-By': 'ASP.NET', 'Date': 'Tue, 15 Jun 2021 09:49:25 GMT'}
-------
301 - http://services.odata.org/V2/Northwind/Northwind.svc/Employees(1)
{'Content-Type': 'text/html; charset=UTF-8', 'Location': 'https://services.odata.org/V2/Northwind/Northwind.svc/Employees(1)', 'Date': 'Tue, 15 Jun 2021 09:49:26 GMT', 'Content-Length': '0'}
-------
200 - https://services.odata.org/V2/Northwind/Northwind.svc/Employees(1)
{'Cache-Control': 'private', 'Content-Length': '18369', 'Content-Type': 'application/json;charset=utf-8', 'Content-Encoding': 'gzip', 'Expires': 'Tue, 15 Jun 2021 09:50:27 GMT', 'Vary': '*', 'Server': 'Microsoft-IIS/10.0', 'DataServiceVersion': '1.0;', 'X-AspNet-Version': '4.0.30319', 'X-Powered-By': 'ASP.NET', 'Date': 'Tue, 15 Jun 2021 09:49:26 GMT'}
-------
Nancy
301 - http://services.odata.org/V2/Northwind/Northwind.svc/Employees(2)
{'Content-Type': 'text/html; charset=UTF-8', 'Location': 'https://services.odata.org/V2/Northwind/Northwind.svc/Employees(2)', 'Date': 'Tue, 15 Jun 2021 09:49:27 GMT', 'Content-Length': '0'}
-------
200 - https://services.odata.org/V2/Northwind/Northwind.svc/Employees(2)
{'Cache-Control': 'private', 'Content-Length': '16074', 'Content-Type': 'application/json;charset=utf-8', 'Content-Encoding': 'gzip', 'Expires': 'Tue, 15 Jun 2021 09:50:27 GMT', 'Vary': '*', 'Server': 'Microsoft-IIS/10.0', 'DataServiceVersion': '1.0;', 'X-AspNet-Version': '4.0.30319', 'X-Powered-By': 'ASP.NET', 'Date': 'Tue, 15 Jun 2021 09:49:26 GMT'}
-------
Andrew

```

phanak-sap (changes requested):
Request change - isolate the 3 lines outside of scope of Observer to new PR.

mnezerka:
@phanak-sap Regarding your sample:
```
employee1 = northwind.entity_sets.Employees.get_entity(1).execute()
print(employee1.FirstName)
employee2 = northwind.entity_sets.Employees.get_entity(2).execute()
print(employee2.FirstName)
```
Are you sure you will be able to pair individual requests with calls of `print_url` hook without putting some glue inside each request? Your example is very simple synchronous scenario. Imagine you are sharing instance of pyodata client (and request session) by more threads or other asynchronous constructions/pools. Your callback will be called, but it could be hard to find originating request for it.

jfilak:
@phanak-sap I asked @mnezerka  to implement this PR. I need to be able to get Headers for some requests. What's your proposal?

docs/conf.py
jfilak: Nah, you update copyright, when you touch the code. This is change is OK.

docs/usage/advanced.rst
jfilak: The only issue with the paragraph is that it mentions "python requests", otherwise I see no untrue parts :)

phanak-sap:
> @phanak-sap Regarding your sample:
> 
> ```
> employee1 = northwind.entity_sets.Employees.get_entity(1).execute()
> print(employee1.FirstName)
> employee2 = northwind.entity_sets.Employees.get_entity(2).execute()
> print(employee2.FirstName)
> ```
> 
> Are you sure you will be able to pair individual requests with calls of `print_url` hook without putting some glue inside each request? Your example is very simple synchronous scenario. Imagine you are sharing instance of pyodata client (and request session) by more threads or other asynchronous constructions/pools. Your callback will be called, but it could be hard to find originating request for it.

Each response instance is calling the hooked function on itself in the Requests lib. I don't understand what kind of gluing you are talking about, but the ` s.hooks['response'].append(print_url)` is only registering the hook on every request the sesssion object made, but the hook function have the 'r' argument to referring the response instance, so the particular response data would be accessible. The originating request is the request URL, printed in the example. Or you need for some purpose the original request variable, in the hook itself you have aceess the [response.request ](https://docs.python-requests.org/en/master/api/#requests.Response.request) instanceinstance.

Source: https://github.com/psf/requests/blob/c45a4dfe6bfc6017d4ea7e9f051d6cc30972b310/requests/sessions.py#L662

I don't see problem for threads or asynchronous usage.

phanak-sap:
> @phanak-sap I asked @mnezerka to implement this PR. I need to be able to get Headers for some requests. What's your proposal?

Well, Point 2 in my comment is the proposed workaround for this use case. Is there anything that is accessible only by this new change and not by the Requests own hook system?

jfilak:
> Well, Point 2 in my comment is the proposed workaround for this use case. Is there anything that is accessible only by this new change and not by the Requests own hook system? If yes, I withdraw my case. If not, I don't see the necessity for the PR.

The point 2 is good, I didn't know about Requests Hooks. However, I though I will change Michal's observer to a decorator, so consumer can do automatic post-processing of retrieved data.

mnezerka:
> > @phanak-sap I asked @mnezerka to implement this PR. I need to be able to get Headers for some requests. What's your proposal?
> 
> Well, Point 2 in my comment is the proposed workaround for this use case. Is there anything that is accessible only by this new change and not by the Requests own hook system? If yes, I withdraw my case. If not, I don't see the necessity for the PR.

If you call:
```
thread1: employee1 = northwind.entity_sets.Employees.get_entity(1).execute()
thread2: employee2 = northwind.entity_sets.Employees.get_entity(2).execute()
```
so you don't know which call was executed first. Request ooks will call your callback twice. How do you know the originating reuqest? I understand that print_url gets instance of request or response, but this instance is from network layer (requests) and not pyodata. And this was my point - will you be able to match it to pyodata requests? So will you be able to sady which one was for employee1 and which for employee2?

docs/usage/advanced.rst
phanak-sap: that was the only line this comment was referring to :)

phanak-sap:
@mnezerka OK, I seems to understand now. Since you are always talking about "hard to find originating request for it" or " pair individual requests with calls of print_url hook" and the Description of the PR is "This change allows caller to observe technical details (e.g. headers, body, etc.) of http requests that are sent in the background of odata calls." it gave me a wrong impression.
I thought you were always talking ONLY about the actual networking layer, that you want to be able to somehow to work differently with the Requests library response instance, based on HTTP status or headers. 

It is true that in the hook workaround, you are outside of scope of the ODataHttpResponse.execute() and you see only the URL for the response, not the protected values in the pyodata scope above it (what Entitiset etc.). So far it never seemed it was a problem, use case seemed to be logging of responses based on status code/error in a header.

But you can have anything in the hook. For the purpose of control flow, there could be for example an exception, so you know if the problem is in the employee1 or employe2 in the runtime. Example modified script, for use case of some say batch read/creation of some entities, which sometimes fails and we want to log what entity was having problem. Observer is another way, but with the necessity of pyodata code change (which I argue is not needed, not that it will not work). 

```
import requests
import pyodata

def print_url(r, *args, **kwargs):
    print(str(r.status_code) + ' - ' + r.url)
    print(r.headers)
    print('-------')

    if r.status_code >= 400:
        raise RuntimeError(r) #most probably some own Exception class

SERVICE_URL = 'http://services.odata.org/V2/Northwind/Northwind.svc/'

s = requests.Session()
s.hooks['response'].append(print_url)

northwind = pyodata.Client(SERVICE_URL, s)

employees = [1,-123, 2]

for id in employees:

    try:
        employee = northwind.entity_sets.Employees.get_entity(id).execute()
        print(employee.FirstName)
    except RuntimeError as e:
        print("id: "+ str(id) + ",error: {0}".format(e))

```

Output:

```
301 - http://services.odata.org/V2/Northwind/Northwind.svc/$metadata
{'Content-Type': 'text/html; charset=UTF-8', 'Location': 'https://services.odata.org/V2/Northwind/Northwind.svc/$metadata', 'Date': 'Tue, 15 Jun 2021 14:58:48 GMT', 'Content-Length': '0'}
-------
200 - https://services.odata.org/V2/Northwind/Northwind.svc/$metadata
{'Cache-Control': 'private', 'Content-Length': '39035', 'Content-Type': 'application/xml;charset=utf-8', 'Expires': 'Tue, 15 Jun 2021 14:58:50 GMT', 'Vary': '*', 'Server': 'Microsoft-IIS/10.0', 'DataServiceVersion': '1.0;', 'X-AspNet-Version': '4.0.30319', 'X-Powered-By': 'ASP.NET', 'Date': 'Tue, 15 Jun 2021 14:58:48 GMT'}
-------
301 - http://services.odata.org/V2/Northwind/Northwind.svc/Employees(1)
{'Content-Type': 'text/html; charset=UTF-8', 'Location': 'https://services.odata.org/V2/Northwind/Northwind.svc/Employees(1)', 'Date': 'Tue, 15 Jun 2021 14:58:49 GMT', 'Content-Length': '0'}
-------
200 - https://services.odata.org/V2/Northwind/Northwind.svc/Employees(1)
{'Cache-Control': 'private', 'Content-Length': '18369', 'Content-Type': 'application/json;charset=utf-8', 'Content-Encoding': 'gzip', 'Expires': 'Tue, 15 Jun 2021 14:58:50 GMT', 'Vary': '*', 'Server': 'Microsoft-IIS/10.0', 'DataServiceVersion': '1.0;', 'X-AspNet-Version': '4.0.30319', 'X-Powered-By': 'ASP.NET', 'Date': 'Tue, 15 Jun 2021 14:58:49 GMT'}
-------
Nancy
301 - http://services.odata.org/V2/Northwind/Northwind.svc/Employees(-123)
{'Content-Type': 'text/html; charset=UTF-8', 'Location': 'https://services.odata.org/V2/Northwind/Northwind.svc/Employees(-123)', 'Date': 'Tue, 15 Jun 2021 14:58:49 GMT', 'Content-Length': '0'}
-------
404 - https://services.odata.org/V2/Northwind/Northwind.svc/Employees(-123)
{'Cache-Control': 'private', 'Content-Length': '125', 'Content-Type': 'application/json', 'Expires': 'Tue, 15 Jun 2021 14:59:50 GMT', 'Vary': '*', 'Server': 'Microsoft-IIS/10.0', 'DataServiceVersion': '1.0;', 'X-AspNet-Version': '4.0.30319', 'X-Powered-By': 'ASP.NET', 'Date': 'Tue, 15 Jun 2021 14:58:50 GMT'}
-------
id: -123,error: <Response [404]>
```

Any other use case which the Observer class must exists? I still think for the practical usage, hooks do just fine.  Async/Threading should work IMHO just fine, each response instance and its hook call is isolated to another, if you think otherwise pls show me sample which fails.

phanak-sap:
> SAP BusinessGateway returns logs in HTTP headers and thought we could imbue the logs to return values of the method execute().

OK, this one looks like is unsolvable by the hook, since it operates above and below the execute() method. Is this still needed or the proposed workaround with throwing the exception solve the same use case?

mnezerka:
> @mnezerka OK, I seems to understand now. Since you are always talking about "hard to find originating request for it" or " pair individual requests with calls of print_url hook" and the Description of the PR is "This change allows caller to observe technical details (e.g. headers, body, etc.) of http requests that are sent in the background of odata calls." it gave me a wrong impression.
> I thought you were always talking ONLY about the actual networking layer, that you want to be able to somehow to work differently with the Requests library response instance, based on HTTP status or headers.
> 
> It is true that in the hook workaround, you are outside of scope of the ODataHttpResponse.execute() and you see only the URL for the response, not the protected values in the pyodata scope above it (what Entitiset etc.). So far it never seemed it was a problem, use case seemed to be logging of responses based on status code/error in a header.
> 
> But you can have anything in the hook. For the purpose of control flow, there could be for example an exception, so you know if the problem is in the employee1 or employe2 in the runtime. Example modified script, for use case of some say batch read/creation of some entities, which sometimes fails and we want to log what entity was having problem. Observer is another way, but with the necessity of pyodata code change (which I argue is not needed).
> 
> ```
> import requests
> import pyodata
> 
> def print_url(r, *args, **kwargs):
>     print(str(r.status_code) + ' - ' + r.url)
>     print(r.headers)
>     print('-------')
> 
>     if r.status_code >= 400:
>         raise RuntimeError(r) #most probably some own Exception class
> 
> SERVICE_URL = 'http://services.odata.org/V2/Northwind/Northwind.svc/'
> 
> s = requests.Session()
> s.hooks['response'].append(print_url)
> 
> northwind = pyodata.Client(SERVICE_URL, s)
> 
> employees = [1,-123, 2]
> 
> for id in employees:
> 
>     try:
>         employee = northwind.entity_sets.Employees.get_entity(id).execute()
>         print(employee.FirstName)
>     except RuntimeError as e:
>         print("id: "+ str(id) + ",error: {0}".format(e))
> ```
> 
> Output:
> 
> ```
> 301 - http://services.odata.org/V2/Northwind/Northwind.svc/$metadata
> {'Content-Type': 'text/html; charset=UTF-8', 'Location': 'https://services.odata.org/V2/Northwind/Northwind.svc/$metadata', 'Date': 'Tue, 15 Jun 2021 14:58:48 GMT', 'Content-Length': '0'}
> -------
> 200 - https://services.odata.org/V2/Northwind/Northwind.svc/$metadata
> {'Cache-Control': 'private', 'Content-Length': '39035', 'Content-Type': 'application/xml;charset=utf-8', 'Expires': 'Tue, 15 Jun 2021 14:58:50 GMT', 'Vary': '*', 'Server': 'Microsoft-IIS/10.0', 'DataServiceVersion': '1.0;', 'X-AspNet-Version': '4.0.30319', 'X-Powered-By': 'ASP.NET', 'Date': 'Tue, 15 Jun 2021 14:58:48 GMT'}
> -------
> 301 - http://services.odata.org/V2/Northwind/Northwind.svc/Employees(1)
> {'Content-Type': 'text/html; charset=UTF-8', 'Location': 'https://services.odata.org/V2/Northwind/Northwind.svc/Employees(1)', 'Date': 'Tue, 15 Jun 2021 14:58:49 GMT', 'Content-Length': '0'}
> -------
> 200 - https://services.odata.org/V2/Northwind/Northwind.svc/Employees(1)
> {'Cache-Control': 'private', 'Content-Length': '18369', 'Content-Type': 'application/json;charset=utf-8', 'Content-Encoding': 'gzip', 'Expires': 'Tue, 15 Jun 2021 14:58:50 GMT', 'Vary': '*', 'Server': 'Microsoft-IIS/10.0', 'DataServiceVersion': '1.0;', 'X-AspNet-Version': '4.0.30319', 'X-Powered-By': 'ASP.NET', 'Date': 'Tue, 15 Jun 2021 14:58:49 GMT'}
> -------
> Nancy
> 301 - http://services.odata.org/V2/Northwind/Northwind.svc/Employees(-123)
> {'Content-Type': 'text/html; charset=UTF-8', 'Location': 'https://services.odata.org/V2/Northwind/Northwind.svc/Employees(-123)', 'Date': 'Tue, 15 Jun 2021 14:58:49 GMT', 'Content-Length': '0'}
> -------
> 404 - https://services.odata.org/V2/Northwind/Northwind.svc/Employees(-123)
> {'Cache-Control': 'private', 'Content-Length': '125', 'Content-Type': 'application/json', 'Expires': 'Tue, 15 Jun 2021 14:59:50 GMT', 'Vary': '*', 'Server': 'Microsoft-IIS/10.0', 'DataServiceVersion': '1.0;', 'X-AspNet-Version': '4.0.30319', 'X-Powered-By': 'ASP.NET', 'Date': 'Tue, 15 Jun 2021 14:58:50 GMT'}
> -------
> id: -123,error: <Response [404]>
> ```
> 
> Any other use case which the Observer class must exists? I still think for the practical usage, hooks do just fine. Async/Threading should work IMHO just fine, each response instance and its hook call is isolated to another, if you think otherwise pls show me sample which fails.

Real use case and also motivation for this pull request:
There are odata services that communicate errors not via HTTP status code, but in content of the one specific header. There is no way how to process such errors smoothly. This PR brings one solution to solve this problem, be backward compatible and stay "user friendly". All you have to to is to check response catched by observer and check for this special header. The important is that your code is sequential, so you can continue or raise error based on response headers catched by observer. I don't see an easy way how to achieve same with hooks - of course, it is possible, but I decided to implement it this way. That's all.

phanak-sap:
Motivation for PR was described perfectly before and it is valid, no discussion about that.

There is no explanation so far why the requests+hook+throw exception if status code or response header == ""gateway error" should not work in threaded or async calls as well. Yes, my example is sequential, it is example of what I mean in the comments by "use requests hooks" in code, since just URL do documentation was not enough. But seems it must be created more complex example. 

So what about modified example with ThreadPoolExecutor and requsts-futures wrapper? The hook workaround works as expected, checking the values of correct response instance.  

```
# pip install requests requests-futures pyodata

import requests
import pyodata
import logging
from requests_futures.sessions import FuturesSession

session = FuturesSession() # By default a ThreadPoolExecutor is created with 8 worker

#logging.basicConfig()
#root_logger = logging.getLogger()
#root_logger.setLevel(logging.DEBUG)

def print_url(r, *args, **kwargs):
    print(str(r.status_code) + ' - ' + r.url)
    print(r.headers)
    print('-------')
    if r.status_code >= 400:
        raise RuntimeError(r)

SERVICE_URL = 'http://services.odata.org/V2/Northwind/Northwind.svc/'

s = requests.Session()
s.hooks['response'].append(print_url)

northwind = pyodata.Client(SERVICE_URL, s)

employees = range(-5,20) #1-9 is returning data, rest is HTTP 404

for id in employees:
    try:
        employee = northwind.entity_sets.Employees.get_entity(id).execute()
        print(employee.FirstName)
    except RuntimeError as e:
        print(e)
```

phanak-sap:
After some investigation, it seems to me that the whole "hooks will not work in async way / Imagine you are sharing instance of pyodata client (and request session) by more threads or other asynchronous constructions/pools" is just mental exercise "what if", while nobody so far really tried it. 

Because it seems it will not work anyway and with Pyodata at the moment you are stuck in sequential requests whether you like it or not.

1. Requests library is blocking in its core.  See [documentation](https://docs.python-requests.org/en/latest/user/advanced/#blocking-or-non-blocking)
2. Alternatives are not working:
    - grequests lib is python 2 only. But you can try gevent and its monkey_patching and wrap the pyodata calls, [it is working here  ](https://github.com/SAP/odfuzz/blob/62b4c17e585305c8864dfb842b75f0832f18f578/odfuzz/fuzzer.py#L251) - but if the requests are blocking the IO, the better performance in my link is gained from asynchronous network calls switching with mongoDB + direct filesystem usage. AFAIK, gevent will not bring anything better to just pyodata + requests stack.
    - requsts-futures will not work with pyodata, fails in `client.py._fetch_metadata ` because method do not expect Futures instance.
    - requests-thread is just experiment, not real working library
    - only httpx looks promising, but I did not tried how well is compatible with requests and pyodata. It will be nice integration test if it works, for sure.
3. Requests session is one per thread anyway, [recommended by the lib authors. ](https://toolbelt.readthedocs.io/en/latest/threading.html)
4. [aiohttp ](https://docs.aiohttp.org/en/v0.22.4/client.html)is the most used non-blocking http lib in python at the moment, but the compatibility with Pyodata is unknown to me and in first glance I guess it is no,  It would be nice integration test - and if the pyodata would not only work with requests, but with aiohttp (and aiohttp would not have similar hook mechanism, did not check), generic Observer pattern mechanism would definitely made sense in such case. 

It would be nice to see the actual working async/multithreading example of pyodata usage - to really prove the hook workaround  is not feasible.

phanak-sap (dismissed):

docs/conf.py
mnezerka: Moved to separate commit to make it easier to cherry pick this change.

docs/usage/advanced.rst
mnezerka: I slightly rephrased this last sentence. Reference to python requests is sill there, but just as an example.

phanak-sap:
Last discussion about this  (offline, internal channels) was ended with statement from @jfilak, that he would like to re-write it differently (if my memory serves me correctly).

Should it be merged as as? Change to draft and commit changes? Dropped?

phanak-sap:
This was long ignored, but it addressed valid use case.

Pyodata meanwhile changed a lot, too many conflicts to actually resolve and async support added. Requests hook system suggested by me is wrong suggestion also, Requests hooks fire at the wrong level, if you need to act on headers before
pyodata's domain handler run.

I am proposing alternative solution for the generic problem under #305.  The SAP specific problem is added to vendor/SAP and any similar problem could be covered by registering own stateless hook. 

This PR will be closed, because it has bug introduced and break intentionally the abstraction layers (python/pyodata - odata wrapper - http networking wrapper)

- model.py unconditional raise KeyError 

  raise KeyError placed outside the loop body fires on every call, including successful ones.

- Observer not thread-safe despite thread-safety being the justification

  RequestObserverLastCall stores mutable state with no locking, so a shared instance under concurrent calls is a race
  condition.

PR #158, opened 2021-07-07
docs: initializing a client for SAP BTP, ABAP env service

s7oev:
solves #109

Issue #109, opened 2020-06-25
Cannot connect to a SAP Cloud Platform OData service


s7oev:
Hello,

First of all, I want to mention that I am aware of the following code:
```
SERVICE_URL = 'https://odata.example.com/Secret.svc'

session = requests.Session()
session.auth = ('MyUser', 'MyPassword')

theservice = pyodata.Client(SERVICE_URL, session)
```

However, I have created an ABAP service following [this guide](https://developers.sap.com/group.abap-env-restful-managed.html) (more specifically, the service is created in step 3, Define and Expose a CDS-Based Travel Data Model). And, I got a service with the following structure:

<some_id>.abap-web.us10.hana.ondemand.com/sap/opu/odata/sap/<service_name>/

[Here's also the exact link](https://3995a331-5b33-4cab-8eeb-9ef5ad61985d.abap-web.us10.hana.ondemand.com/sap/opu/odata/sap/ZUI_SS_C_TRAVEL_M/) if that's necessary.

I cannot find any way to connect with it through PyOData. I imagine this is because SAP Cloud Platform uses OAuth2, while the example code above - HTTP Basic Auth. Still, I tried to input user and password data just to try, but that didn't work. I also managed to get an authorization token (from an xssec security context), which I think it might help, but can't see how to pass it properly.

Are there any suggestions for connecting to this service?

filak-sap:
I didn't try it but you basically need to do OAuth2 authorization as described at:
https://requests-oauthlib.readthedocs.io/en/latest/oauth2_workflow.html#web-application-flow

and the use the variable oauth (in the 3rd step) instead for the variable session (from the code snippet).

Please, let me know if it works.

s7oev:
Hey, Filak, thanks a lot for your response and suggestions.

Yes, I also tried using OAuth2Session already, tried again, unfortunately without luck. Best result I got was with this code:
```
from requests_oauthlib import OAuth2Session

client_id = r'{id here}'
client_secret = r'{secret here}'
redirect_uri = 'https://myapp778.cfapps.us10.hana.ondemand.com/sap/opu/odata/sap/ZUI_SS_C_TRAVEL_M/'

oauth = OAuth2Session(client_id, redirect_uri=redirect_uri,)
authorization_url, state = oauth.authorization_url('https://myapp778.cfapps.us10.hana.ondemand.com/')
print('Please go to %s and authorize access.' % authorization_url)

authorization_response = input('Enter the full callback URL')
[...]

PR #159, opened 2021-07-11
vendors: token helper for SAP BTP, ABAP env service

s7oev:
solves #109, rewritten to helper as requested in #158

@jfilak , I think functionality-wise this is finished (including tests); I could add some docs but not sure what would be best location - does it make sense to put them in initialization, or best to make a new, separate, top-level section `vendors.rst`?

also a comment for the implementation: `add_btp_token_to_session()` requires the session object to be given as argument, instead of creating it inside - this allows to avoid adding a `requests` dependency. so the intended usage is something like this:

```python
session = SAP.add_btp_token_to_session(requests.Session(), KEY, USER, PASSWORD)
# do something more with session object if necessary (e.g. adding sap-client parameter, or CSRF token)
client = pyodata.Client(SERVICE_URL, session)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+- Helper for obtaining a token for services coming from SAP BTP, ABAP environment - Stoyko Stoev
+
 ## [1.7.0]
 
 ### Added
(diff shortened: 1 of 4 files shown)


filak-sap:
Thank you very much. I like it.
Putting the documentation into a solo file is a good idea - I will leave the PR open until you add it.

phanak-sap:
If you will create doc file to this PR, you can commit update of [changelog.md](https://github.com/SAP/python-pyodata/blob/master/CHANGELOG.md)

s7oev:
@filak-sap @phanak-sap 
there's a new commit with both things done

filak-sap:
Well done!

Issue #165, opened 2021-07-24
Getting an EntitySetProxy


avoutsas67:
Hello, 

I would like reference dynamically ODATA entities. As per the tutorial one way of getting how many records an entity_set has is: 
`service.entity_sets.entity_set_name.get_entities().count().execute()`

I would like to pass the name of the entity set as a parameter and get the number of records in the given data set.  E.g.
`service.schema.entity_set('entity_set_name') `

However the above code returns and EntitySet which does not have the method get_entities(). 
Can you please explain what is the most appropriate method of getting an EntitySetProxy entity passing the name of the entity set as a parameter? 

Many thanks in advance for your help,

Achilleas

phanak-sap:
Hi @avoutsas67 

First, `service.schema.entity_set` is traversing the service $metadata, so it make sense to not have `get_entities` there. It just gives you the service structure. 

Second, it seems to me that referencing odata entities dynamically is not covered in exposed API. Seems as enhancement request.

However, I think I was able to get the information you seek in a ugly way, by using the Python magic methods, e.g. 
`sample.entity_sets.__getattr__('entity_set_name').get_entities().count().execute()`

Is that acceptable workaround for you?

avoutsas67:
Dear @phanak-sap,

Many thanks for the suggestion. The following code
`service.entity_sets.__getattr__('entity_set_name').get_entities().count().execute()`
does the job

Thanks for your response and prompt action. Providing an example in the documentation would be really welcome.

Many thanks for your support.

Kind regards

Achilleas

PR #167, opened 2021-07-28
Update advanced.rst

dmatula:
doc: Dynamically referenced EntitySet  for issue #165

### DIFF
--- a/docs/usage/advanced.rst
+++ b/docs/usage/advanced.rst
@@ -96,10 +96,10 @@ it is enough to set log level to the desired value.
     root_logger = logging.getLogger()
     root_logger.setLevel(logging.DEBUG)
     
-Get a count of entities using a getattr() function
+Dynamically referenced EntitySet
 ------------------------------------------------
 
-Print a count of all employees with getattr():
+If you need to work with many Entity Sets the same way or if you just need to pick up the used Entity Set name in run-time, you may find out the ability to get an instance of Entity Set proxy dynamically. Here is an example of how you can print a count of all employees:
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


docs/usage/advanced.rst
```diff
@@ -95,3 +95,13 @@ it is enough to set log level to the desired value.
     logging.basicConfig()
     root_logger = logging.getLogger()
     root_logger.setLevel(logging.DEBUG)
+    
+Get a count of entities using a getattr() function
```
phanak-sap, line 99: This is too specific and in wrong doc section. 

This example belongs to the basic querying doc page - for example directly after first subsection. Ignore the count since it is over-specific example, the main focus is the getattr() usage.

Since this is referencing the issue Getting an EntitySetProxy #165, the intent of the user, in similar fashion described there is better as subtitle - e.g. "Get one entity referenced dynamically" or similar

phanak-sap (changes requested):

docs/usage/advanced.rst
jfilak: I agree with too specific title -> the title does not describe what was achieved, it describes the technical solution but users in the need for the solution does not know the technical solution so they cannot use the title.

I do not  agree with the objection about a wrong section as this definitely belongs to the Advanced section and the querying is used just to show what can be done. The title could be "Dynamically referenced EntitySet" ?

phanak-sap: OK, I was seeing it as simple usage of one of [python built-ins](https://docs.python.org/3/library/functions.html) for something pyodata has no fluent API - just the addressing of the entitysets like in the first query user will ever read in docs; querying.rst the first "howto practicaly use pyodata" after "how to initialize". Everything in advanced (Batch, Errors, Logging) are topics unrelated to each other and something that users may want to read after their first working queries - s so that was my rationale. 

But I agree the second point is not that important as the first about the too specific title and can be in both .rst files; just that there is something to reference in future issues :)

filak-sap:
Thank you!

PR #175, opened 2021-08-20
Always escape special chars in URL query params

simonbru:
Some special URL characters (such as `+`, `=` and `&`) were not escaped in query parameters. This would introduce a risk of data corruption or URL injection.

Note: I did not notice any breakage when testing against SAP. However, I only tested a few use cases (e.g., I barely tested batched requests).

### DIFF
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -13,13 +13,8 @@ import random
 from email.parser import Parser
 from http.client import HTTPResponse
 from io import BytesIO
+from urllib.parse import urlencode
 
-try:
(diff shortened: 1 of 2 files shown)


filak-sap:
Good catch! Thank you very much. Can you accept the CLA, please? I cannot merge your patch without it.

simonbru:
>Can you accept the CLA, please?
Done !

phanak-sap:
CLA signed, PR looking good. I will update changelog in follow-up commit.

simonbru:
Thanks for the quick response!

PR #183, opened 2021-12-13
model: split collections in Schema class

phanak-sap:
model: split properties schema.entity_types/complex_types and their generated Collections 

Changed:
- data from  Schema.entity_sets are now in Schema.complex_types and Schema._collections_entity_sets 
- data from  Schema.complex_types are now in Schema.complex_types and Schema._collections_complex_types --
- update tests to better cover generated Collections inside schema parsing

Additional details:
- class Types is still containing both types and their generated collection variant, since it is static class for odata primitive types

Long explanation: 

Collections of entity types were added with focus of "get function imports working", which is main use case of pyodata. See commit 009198379 for details.

For use case of metadata validation, it is however not API user friendly to return two, different classes in one property, and also things that does not exists in the metadata XML at all. Collection of entity type is not "Entity Type". Schema class should return for entity_type property only parsed <EntityType> elements, not with generated possible return types for another elements. Otherwise it is misleading API to user, in a sense that there are things in schema.entity_type that are not present in the XML.

```
for et in service.schema.entity_types:
     print(et.name) 

EntityType(MasterEntity)
Collection(EntityType(MasterEntity))
EntityType(DataEntity)
Collection(EntityType(DataEntity))
```

Workaround for grabbing only entity types exist, but it is ugly and requires from API user deeper understanding than just "I want `schema.entity_sets` and `schema.entity_types` return similar list of respective <things> " .

```
from pyodata.v2.model import Collection 

        for et in self.pyodata.schema.entity_types:
            if not isinstance(et, Collection):
                print(et.name)
```

### DIFF
--- a/pyodata/v2/model.py
+++ b/pyodata/v2/model.py
@@ -805,12 +805,6 @@ class Schema:
         def list_association_sets(self):
             return list(self.association_sets.values())
 
-        def list_collections_entity_types(self):
-            return list(self._collections_entity_types.values())
-
-        def list_collections_complex_types(self):
-            return list(self._collections_complex_types.values())
-
         def add_entity_type(self, etype):
             """Add new  type to the type repository as well as its collection variant"""
 
--- a/pyodata/v2/model.py
+++ b/pyodata/v2/model.py
@@ -782,7 +782,7 @@ class Schema:
 
             # generated collections for ease of lookup (e.g. function import return type)
             self._collections_entity_types = dict()
-            self._collections_complex_types = dict()  # TODO
+            self._collections_complex_types = dict()
 
         def list_entity_types(self):
             return list(self.entity_types.values())
@@ -805,8 +805,11 @@ class Schema:
         def list_association_sets(self):
             return list(self.association_sets.values())
 
-        def list__collections_entity_types(self):
+        def list_collections_entity_types(self):
             return list(self._collections_entity_types.values())
+
+        def list_collections_complex_types(self):
+            return list(self._collections_complex_types.values())
 
         def add_entity_type(self, etype):
             """Add new  type to the type repository as well as its collection variant"""
@@ -828,9 +831,9 @@ class Schema:
             # automatically create and register collection variant if not exists
             if isinstance(ctype, NullType):
                 return
-
             collection_type_name = f'Collection({ctype.name})'
-            self.complex_types[collection_type_name] = Collection(ctype.name, ctype)
+            self._collections_complex_types[collection_type_name] = Collection(ctype.name, ctype)
+            # TODO performance memory: this is generating collection for every entity type encoutered, regardless of such collection is really used.
 
         def add_enum_type(self, etype):
             """Add new enum type to the type repository"""
@@ -874,7 +877,7 @@ class Schema:
         """Returns either EntityType, ComplexType or EnumType that matches the name.
         """
 
-        for type_space in (self.entity_type, self.complex_type, self.enum_type):
+        for type_space in (self.entity_type, self._collections_entity_types, self.complex_type, self._collections_complex_types, self.enum_type):
             try:
                 return type_space(type_name, namespace=namespace)
             except KeyError:
@@ -903,15 +906,15 @@ class Schema:
             try:
                 return self._decls[namespace]._collections_entity_types[type_name]
             except KeyError:
-                raise KeyError(f'EntityType Collection {type_name} does not exist in Schema Namespace {namespace}')
+                raise KeyError(f'EntityType collection {type_name} does not exist in Schema Namespace {namespace}')
 
         for decl in list(self._decls.values()):
             try:
-                return decl.entity_types[type_name]
+                return decl._collections_entity_types[type_name]
             except KeyError:
                 pass
 
-        raise KeyError(f'EntityType {type_name} does not exist in any Schema Namespace')
+        raise KeyError(f'EntityType collection {type_name} does not exist in any Schema Namespace')
 
     def complex_type(self, type_name, namespace=None):
         if namespace is not None:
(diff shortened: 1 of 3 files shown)


pyodata/v2/model.py
```diff
@@ -175,15 +175,14 @@ def parse(value):
...
-    """Repository of all available OData types
+    """Repository of all available OData V2 primitive types + their Collection variants
 
        Since each type has instance of appropriate type, this
        repository acts as central storage for all instances. The
        rule is: don't create any type instances if not necessary,
        always reuse existing instances if possible
     """
 
-    # dictionary of all registered types (primitive, complex and collection variants)
```
phanak-sap, line 186: Maybe this was intent in the start, but now is IMHO misleading.

pyodata/v2/model.py
```diff
@@ -781,6 +780,10 @@ def __init__(self, namespace):
             self.associations = dict()
             self.association_sets = dict()
 
+            #generated collections for ease of lookup (e.g. function import return type)
+            self._collections_entity_types = dict()
+            self._collections_complex_types = dict() #TODO
```
phanak-sap, line 785: This will be in second commit, so first draft is shorter to read. Also, I am modifying relevant test, so should be in separate commits anyway.

tests/test_model_v2.py
```diff
@@ -349,6 +349,9 @@ def test_edmx_complex_types(schema):
     assert str(real_prop) == 'StructTypeProperty(Real)'
     assert str(real_prop.struct_type) == 'ComplexType(ComplexNumber)'
 
+    #after correct parsing, new complex type is registered in metadata schema
+    assert str(schema.typ('ComplexNumber')) == 'ComplexType(ComplexNumber)'
```
phanak-sap, line 353: Added before I start splitting Collations for Complex Types as well, this part of Schema class was not tested at all. 

In effect, if you comment out this line, 120 tests fails.

https://github.com/SAP/python-pyodata/blob/21a99b50c3b1718e5b83109d6b4d944d06b7d1e5/pyodata/v2/model.py#L815

But if you comment out this line, 100% tests pass:

https://github.com/SAP/python-pyodata/blob/21a99b50c3b1718e5b83109d6b4d944d06b7d1e5/pyodata/v2/model.py#L827

phanak-sap:
Open question - should this Collections split  relevant to Annotations as well? I so far looked at schema.entity.types and schema.complex_types only. I am not sure about the annotations, pls review - my basic understanding so far is that this is not related to generated Collections in schema. 

Example: https://github.com/SAP/python-pyodata/blob/21a99b50c3b1718e5b83109d6b4d944d06b7d1e5/tests/test_model_v2.py#L815-L825

phanak-sap:
Note: builds are failing due linter. I run pytest repeatedly so everything is passing and forgot about pylint in the end.

pyodata/v2/model.py
```diff
@@ -802,6 +805,9 @@ def list_associations(self):
         def list_association_sets(self):
             return list(self.association_sets.values())
 
+        def list__collections_entity_types(self):
```
filak-sap, line 808: That double _ is not necessary, use just 1.

pyodata/v2/model.py
filak-sap: Not sure which second commit you have on mind. The standard approach is to just not add what's not necessary.

pyodata/v2/model.py
filak-sap: Reviewing your own changes? Better to put your comments to commit message then.

filak-sap:
I am sorry but I see no tests for the new functionality.

pyodata/v2/model.py
phanak-sap: typo... yeah

phanak-sap:
> I am sorry but I see no tests for the new functionality.

I thought thate existing tests are covering this quite sufficiently - see comment for tests/test_model_v2.py above. But I encoutered same problem mentioned in this comment, fixed in latest commit f86cd1f439. I am adding equivalence check for getting the element trough schema.typ generic method and the new private _collections_XXX method; I have noticed that complex type failures mentioned code path that was not covered in the 195 passing tests - trying to get the Collection(Something) and not just plain entity_type or complex_type element present in XML.

Everything changed / added should be AFAIK called trough these two.

phanak-sap:
This is ready for review and merging to mainline. I am expecting Squashing into one commit with the commit message same as the PR description - feel free to update directly.

filak-sap:
@phanak-sap  codecov claims you have added 4 statements that are not covered by tests. Please, have a look at test coverage and cover those 4 statements if feasible.

phanak-sap:
> @phanak-sap codecov claims you have added 4 statements that are not covered by tests. Please, have a look at test coverage and cover those 4 statements if feasible.

I do not understand codecov:
- it says +4 here, while +2 statements on the[ model.py ](https://app.codecov.io/gh/SAP/python-pyodata/compare/183/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%20comments&utm_term=None#diff-cHlvZGF0YS92Mi9tb2RlbC5weQ==)
- it run IMHO for half of commits in this PR, I am not able to find a way to trigger it again (like travis build)
- it returns different numbers than pytest-cov, which is the only one I am able to run localy with same results as on Travis
```
Name                                    Stmts   Miss  Cover
------------------------------------------------
pyodata/v2/model.py           1574    116    93%
```


- the not covered are functions list_collections_entity_types() / list_collections_complex_types(), which are trivial and no need to write test for these. In fact, I do not know why they are in the Schema anyway, such one-line helper can create any pyodata user and IMHO they are not used anywhere internally, I just wanted to be consistent with the rest of the class.

filak-sap:
```bash
make report-coverage-html
```

phanak-sap:
> ```shell
> make report-coverage-html
> ```

does not seem to do a thing.

I did removed the two new functions, I thought that they will be useful on their own somehow like the rest, did not truly check where they are used - properties of Schema. The point of this PR is that generated collections should not be part of "module API", so I am having to problem to just delete the code in the name of better coverage.

pyodata/v2/model.py
phanak-sap: I know.. But it was marked as WIP PR,.. and I did not want to do all the necessary work, only to be rejected... so wanted some preliminary review if the idea is valid half-way.

phanak-sap:
Is there anything that should be addressed? As I see it, this is ready to be merged to master, just don't want to do that for my own code.

filak-sap:
Thank you! Good job!

phanak-sap:
This was ment to be squashed to one, cleaner commit. I would do that if the PR was just marked as Approved.

filak-sap:
> This was ment to be squashed to one, cleaner commit. I would do that if the PR was just marked as Approved.

I am an oldskool maintainer.
I keep the PR clean and I expect the reviewer merges.
I squashed you random commits and force pushed master -> https://github.com/SAP/python-pyodata/commit/bfa215f138d7e76737fb6d2e268a513e30a7ba84

PR #184, opened 2021-12-22
model: fix timezone handling for Edm.DateTimeOffset

rettichschnidi:
Convert Edm.DateTimeOffset entities to Python datetime objects with a
timezone attached, instead of simply passing them around as plain
string.


phanak-sap:
Seems that [current_timezone()](https://github.com/SAP/python-pyodata/blob/21a99b50c3b1718e5b83109d6b4d944d06b7d1e5/pyodata/v2/model.py#L31) method (which should probably be deleted in the PR as well, since this replaces its usages) and the comments there are wrong.

But then, this same problem should be fixed for Edm.DateTime as well, not only for Edm.DateTimeOffset. Correct?

Note, relevant documents: 

- Spec in [odata.org](https://www.odata.org/documentation/odata-version-2-0/overview/#AbstractTypeSystem) - Edm.DateTimeOffset: Defined by the lexical representation for datetime (including timezone offset)
- [MS](https://docs.microsoft.com/en-us/openspecs/windows_protocols/mc-csdl/352cece0-c8b9-4e21-b258-170b77900502)
- [XML schema, no anchors]( https://www.w3.org/TR/xmlschema-2/) - Section 3.2.7.3 Timezones

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
```
phanak-sap, line 9: This is not new feature, but bugfix (DateTime is already supported, however expects that everything is in UTC). 

Pls change from "Added" to "Fixed" section, as we are using https://keepachangelog.com/en/1.0.0/

pyodata/v2/model.py
```diff
@@ -361,6 +362,37 @@ def from_literal(self, value):
         return matches.group(1)
 
 
+def ms_since_epoch_to_datetime(value, tzinfo):
```
phanak-sap, line 365: what the "ms" prefix stands for?

tests/test_model_v2.py
```diff
@@ -596,6 +596,150 @@ def test_traits_datetime():
     assert str(e_info.value).startswith('Cannot decode datetime from value xyz')
 
 
+def test_traits_datetimeoffset():
```
phanak-sap, line 599: Rather then one big test with comments like " # 1. direction Python -> OData" it would be better to simply just split to multiple isolated tests anyway..

tests/test_model_v2.py
```diff
@@ -596,6 +596,150 @@ def test_traits_datetime():
...
+    assert testdate.tzinfo == timezone(-timedelta(hours=1))
+
+    # UTC
+    testdate = typ.traits.from_literal("datetimeoffset'1976-11-23T03:33:06Z'")
+    assert testdate.year == 1976
+    assert testdate.second == 6
+    assert testdate.microsecond == 0
+    assert testdate.tzinfo == timezone.utc
+
+    # UTC lowercase
```
phanak-sap, line 646: where us the lowercase 'z' test case coming from? XML schema defines only 'Z' - see section 3.2.7.3 Timezones, cite:

_The lexical representation of a timezone is a string of the form: (('+' | '-') hh ':' mm) | 'Z',_  

On the other hand, for the UTC timezone, two other offset test cases are also valid input that should end to same result:  `+00:00` and `-00:00 `

pyodata/v2/model.py
rettichschnidi: milliseconds - but that made the function name too long (>30 characters), and pylint complained. Will add a comment.

CHANGELOG.md
rettichschnidi: > Seems that [current_timezone()](https://github.com/SAP/python-pyodata/blob/21a99b50c3b1718e5b83109d6b4d944d06b7d1e5/pyodata/v2/model.py#L31) method (which should probably be deleted in the PR as well, since this replaces its usages) and the comments there are wrong.

Not sure I understand you correctly, but this would essentially mean that Edm.DateTime will be converted to [naive datetime](https://docs.python.org/3/library/datetime.html) objects, instead of "aware" ones with a hard coded UTC timezone?

> But then, this same problem should be fixed for Edm.DateTime as well, not only for Edm.DateTimeOffset. Correct?

The literal form of Edm.DateTime in (Overview)[https://www.odata.org/documentation/odata-version-2-0/overview/#AbstractTypeSystem) does not mention either, timezone nor offset. [Atom Format](https://www.odata.org/documentation/odata-version-2-0/atom-format/) references the literal form directly, therefore can also not represent a timezone/offset. [JSON](https://www.odata.org/documentation/odata-version-2-0/json-format/) however has an offset described. :confused:

As I have yet not seen a timezone/offset on a Edm.DateTime, I assumed it will never have an offset. Additionally, what would be the purpose of Edm.DateTimeOffset if both types can contain a timezone?

Possible way forward:
- Edm.DateTime: Remove fake timezone information (separate PR?)
- Edm.DateTimeOffset: Add timezone information (original idea of this PR)

tests/test_model_v2.py
rettichschnidi: Will do

CHANGELOG.md
rettichschnidi: > This is not new feature, but bugfix (DateTime and DateTimeOffset are already supported, however they expects that everything incoming is in UTC).

The level of support currently differs:
- DateTime: Converted to a Python datetime object with fake/guessed UTC timezone
- DateTimeOffset: Converted to Python string. Works well enough to fetch and send to services, but not when one wants to perform work on it. This PR converts to/from a Python datetime object with the proper timezone attached.

> Pls change from "Added" to "Fixed" section, as we are using https://keepachangelog.com/en/1.0.0/

Will do.

tests/test_model_v2.py
rettichschnidi: Error on my side - while datetime is case-insensitive as per spec, this is not the case for the time zone part.

Will add the suggested tests.

rettichschnidi:
Just pushed with the requested changes that where clear to me.

CHANGELOG.md
phanak-sap: Edm.DateTime - while odata.org has is just datetime’yyyy-mm-ddThh:mm[:ss[.fffffff]]’ - and SAP Gateway returns this in UTC by design, so it is understandable that pyodata started from there - the MS ODATA specification for odata has optional timezone as well for this. That's why I thought that what you did for Edm.DateTimeOffset type should be applied to Edm.DateTime as well - and apply UTC as the default, if no infomration about timezone is received. 

                 
```
dateTimeLiteral =   year "-" 
                    month "-" 
                    day "T"
                    hour ":" 
                    minute 
                    [":" second ["." nanoSeconds]] 
                    [timeZone]
```

https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-odata/61934eae-311a-4af4-b8f8-82c112248651

pyodata/v2/model.py
phanak-sap: Oh, obviously... anyway, docstring will help :)

tests/test_model_v2.py
phanak-sap: If we have some tests like that, than shame on us as well and we should rewrite that to shorter, isolated test cases :)

tests/test_model_v2.py
phanak-sap: One other funny test cases... add something on the dateline or even across - +/- 12h timezone exactly or  +14 (Kiribati), so even the date should change accordingly after the conversion.

pyodata/v2/model.py
```diff
@@ -361,6 +362,38 @@ def from_literal(self, value):
...
+    """Convert milliseconds since midnight 1.1.1970 to datetime"""
+    try:
+        # https://stackoverflow.com/questions/36179914/timestamp-out-of-range-for-platform-localtime-gmtime-function
+        return datetime.datetime(1970, 1, 1, tzinfo=tzinfo) + datetime.timedelta(milliseconds=int(value))
+    except (ValueError, OverflowError):
+        if FIX_SCREWED_UP_MINIMAL_DATETIME_VALUE and int(value) < -62135596800000:
```
phanak-sap, line 371: The -62135596800000 and 253402300799999 are obviously the "magic numbers", that repeats in if statement and the log error message as well. Local constant in the method would be nicer.

you  see the statement yourself:
` if FIX_SCREWED_UP_MAXIMUM_DATETIME_VALUE and int(value) > 253402300799999:`

pyodata/v2/model.py
```diff
@@ -361,6 +362,38 @@ def from_literal(self, value):
...
+            return datetime.datetime(year=1, day=1, month=1, tzinfo=tzinfo)
+        if FIX_SCREWED_UP_MAXIMUM_DATETIME_VALUE and int(value) > 253402300799999:
+            return datetime.datetime(year=9999, day=31, month=12, tzinfo=tzinfo)
+        raise PyODataModelError(f'Cannot decode datetime from value {value}. '
+                                f'Possible value range: -62135596800000 to 253402300799999. '
+                                f'You may fix this by setting `FIX_SCREWED_UP_MINIMAL_DATETIME_VALUE` '
+                                f' or `FIX_SCREWED_UP_MAXIMUM_DATETIME_VALUE` as a workaround.')
+
+
+def parse_datetime_literal(value):
```
phanak-sap, line 384: 1) For future generations, this should have short docstring why these three formats were chosen. I did not check to be fair, but I am guessing all statements and possible code path are covered by new tests, right?

2)
This looks to me like ideal example for usage [pytest parametrized test](https://docs.pytest.org/en/6.2.x/parametrize.html)s, with throwing bunch of valid inputs and invalid inputs to two tests, with difference in expected result. The collections for valid a invalid input string can be in conftest if too big to have the test readable.

This is idea how to possibly enhance the PR, not a blocker to merging (will convert to issue for myself if not done)

pyodata/v2/model.py
```diff
@@ -439,18 +455,87 @@ def from_literal(self, value):
...
+       datetimeoffset'yyyy-mm-ddThh:mm[:ss]±ii:nn' (works for all time zones)
+       datetimeoffset'yyyy-mm-ddThh:mm[:ss]Z' (works only for UTC)
+       NOTE: Spaces are not allowed between datetimeoffset and quoted portion.
+       The datetime part is case-insensitive, the offset one is not.
+
+       Example 1: datetimeoffset'1970-01-01T00:00:01+00:30'
+        - /Date(1000+0030)/ (As DateTime, but with a 30 minutes timezone offset)
+       Example 1: datetimeoffset'1970-01-01T00:00:01-00:60'
+        - /Date(1000-0030)/ (As DateTime, but with a negative 60 minutes timezone offset)
+       https://blogs.sap.com/2017/01/05/date-and-time-in-sap-gateway-foundation/
```
phanak-sap, line 478: blog SAP is sadly not the greatest reference. I used links to odata spec in this issue and in #186 - in fact, we found that what the SAP Gateway do, is kind of "odata flavor" and should not be considered same plain, vanilla odata specification.

pyodata/v2/model.py
```diff
@@ -439,18 +455,87 @@ def from_literal(self, value):
...
+        if isinstance(value, str):
+            return value
+
+        return f'/Date({int(value.timestamp()) * 1000}{value.utcoffset(value):+})/'
+
+    def from_json(self, value):
+        if value is None:
+            return None
+
+        matches = re.match(r"^/Date\((?P<milliseconds_since_epoch>-?\d+)(?P<offset_in_minutes>[+-]\d+)\)/$", value)
```
phanak-sap, line 503: I am just asking.. not even want to search to prove a point - regexp instead of something from datetime package is really the only option for parsing datetimeoffset? Just that the line itself looks like many tests should be covering this - I know we have re.match usages in the package as well, just that with regexp for entire line I recollect the old joke "now I have X+1 problems" :)

pyodata/v2/model.py
```diff
@@ -439,18 +455,87 @@ def from_literal(self, value):
...
-            except ValueError:
-                try:
-                    value = datetime.datetime.strptime(value, '%Y-%m-%dT%H:%M')
-                except ValueError:
-                    raise PyODataModelError(f'Cannot decode datetime from value {value}.')
+            milliseconds_since_epoch = matches.group('milliseconds_since_epoch')
+            offset_in_minutes = int(matches.group('offset_in_minutes'))
+        except (ValueError, AttributeError):
+            raise PyODataModelError(
+                f"Malformed value {value} for primitive Edm type. Expected format is /Date(datetime±offset_in_min)/")
```
phanak-sap, line 509: "for primitive Edm type" -> "for primitive Edm.DateTimeOffset". We know where we are fore the log message.

pyodata/v2/model.py
```diff
@@ -439,18 +455,87 @@ def from_literal(self, value):
...
 
-        return value.replace(tzinfo=current_timezone())
+        if value is None:
+            return None
+
+        value = super(EdmDateTimeOffsetTypTraits, self).from_literal(value)
+
+        try:
+            # Note: parse_datetime_literal raises a PyODataModelError exception on invalid formats
+            if re.match(r'\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z', value, flags=re.ASCII | re.IGNORECASE):
```
phanak-sap, line 524: I will copypaste this... I am just asking.. regexp also the only option?

CHANGELOG.md
rettichschnidi: > Edm.DateTime - while odata.org has is just datetime’yyyy-mm-ddThh:mm[:ss[.fffffff]]’ - and SAP Gateway returns this in UTC by design

It will never include any timezone information, correct?

If so: Since Pythons datetime objects can be naive (explicitly not knowing anything about timezone) or aware (having some kind of timezone information), what about working with this to keep track whether a value is timezone aware or not instead of adding a fake UTC timezone?

> the MS ODATA specification for odata has optional timezone as well for this.

This refers to [version 3.0 of  OData](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-odata/c8490c34-adbc-43ee-a8c6-89760c123484). I think it should not be used as reference for OData 2.0.

> That's why I thought that what you did for Edm.DateTimeOffset type should be applied to Edm.DateTime as well - and apply UTC as the default, if no infomration about timezone is received.

Changing Edm.DateTime and Edm.DateTimeOffset so that the only change will be that the first one defaults to UTZ and the later one will fail if the offset/timezone information is missing should be easy to be implemented when /consuming/ data.

However, when /sending/ an aware (possibly non-UTC) Python datetime object to a server (converting it to its literal form) however, I can see two options:
1. Just expecting the server to deal with the timezone information. If there are implementations which can not handle it, hopefully someone will create a bug report.
2. "Normalize" the time information to UTZ. Implies that receiving entities, and then using them (unmodified) to "update" data on the server strips the previously existing timezone information. Feels wrong.

I think both attempts are not sound and Edm.DateTime should be strictly agnostic of timezones, while Edm.DateTimeOffset requires it.

In case the SAP Gateway never adds a timezone information to Edm.DateTime, this approach should work well with it.

pyodata/v2/model.py
rettichschnidi: Not a big fan either, but the only alternatives I could think of are:
- Something ["scanf()"](https://docs.python.org/3/library/re.html#simulating-scanf) based wrapped in some try or if-else clauses
- Thought of but not working: [strftime](https://docs.python.org/3/library/time.html#time.strftime)

pyodata/v2/model.py
rettichschnidi: will fix

pyodata/v2/model.py
rettichschnidi: Alternatives:
- [fromisoformat](https://docs.python.org/3/library/datetime.html#datetime.date.fromisoformat) - but this needs Python 3.7+. Ubuntu 18.04 will be EOL soon - any chance you will drop support 3.6?
- [python-dateutil](https://dateutil.readthedocs.io/en/stable/) - would be a new dependency

pyodata/v2/model.py
rettichschnidi: Given this is a SAP project, changes should remain compatible to the SAP flavor anyway?

pyodata/v2/model.py
rettichschnidi: > 1. For future generations, this should have short docstring why these three formats were chosen.

Will do

> I did not check to be fair, but I am guessing all statements and possible code path are covered by new tests, right?

It gets tested implicitly (but fully) by test_traits_datetime(). test_traits_datetimeoffset_from_literal() only uses the first to forms.

>     2. This looks to me like ideal example for usage [pytest parametrized test](https://docs.pytest.org/en/6.2.x/parametrize.html)s, with throwing bunch of valid inputs and invalid inputs to two tests, with difference in expected result. The collections for valid a invalid input string can be in conftest if too big to have the test readable.

Will add some.

CHANGELOG.md
phanak-sap: OK. I understand this long crossed comment as that the request for covering timezone support for Edm.DateTime is accepted. If it is true, thanks for that. I will leave this PR open.

I would like only to address the odata v3 part in the comment - yes, the MS documentation is quite a chaotic one. They version the document itself, not by the protocol version. The OASIS is clearer for understanding. However, IMHO the document is essentialy "everything is applied to both odata v2 and odata v3" - and things that are relevant to one protocol version only are listed  in the section 1.7.1 and 1.7.2 - https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-odata/31a103dd-acaa-4e3d-a637-2d5de28a20a1

pyodata/v2/model.py
phanak-sap: Great, thank you for that.

pyodata/v2/model.py
phanak-sap: **to clear the possibility of misunderstanding:**

Depends on what exactly you mean by "compatible to the SAP flavor ". Compatible with SAP Gateway, if it produces something different that exact odata specification? Yes. (I know about added annotations for UI). Use the SAP Gateway as the base for everything - since so far changes are in the model.py - no.

Even that it started from within the SAP (so started de-facto with premise of equivalence that SAP Gateway == OData) and is obviously used in SAP, the pyodata as generic python package should be ideally covering odata specification in its vanilla form - and add different vendor flavors support where they differs.  Our[ "sister JS library" ](https://github.com/SAP/odata-library)(which was created from scratch basically as port of this to JS) has this handled better, [in more generic way](https://github.com/SAP/odata-library/tree/985a355a7532f5ba12fdf5a63cf604cde875a57d/lib/model/nw/extensions), but so far there was no other need than vanilla oasis and sap flavor, so we are good :). 

There is is something in [SAP vendor module](https://github.com/SAP/python-pyodata/blob/master/pyodata/vendor/SAP.py), so far handling specific gateway errors. Some SAP specific things from model.py should end there, e.g. https://github.com/SAP/python-pyodata/blob/21a99b50c3b1718e5b83109d6b4d944d06b7d1e5/pyodata/v2/model.py#L1613

Did you find any difference in behavior of timezones that would be SAP specific - and not applicaple to "odata model in general" in model.py? AFAIK, all your changes so far should be odata-generic, without any need for SAP special handling (i.e. SAP adhers to DateTime and DateTimeOffset odata specification).

**actual line comment relevant:**

Everything previous is little tangential, I only commented that I would use the link to odata specification (as scattered in comments in this PR) instead of SAP blog post, that's all.

pyodata/v2/model.py
phanak-sap: Thanks for explanation.

pyodata/v2/model.py
phanak-sap: Thanks for explanation of the thoughts behind this line.

Actually, since you asked.. yes, I have intention in near future drop support for Python 3..6, since it is EOL, will not receive any security fixes - and pyodata is a type of library that most probably will be used in more fresh environments, simply since you want the latest requests library as well, I guess.

So far there was no actual need for code change that would be Py3.6 incompatible - so I wanted to mark 3.6 python build as possible to fail, but monitor them 

I would go in a way that we create an spinoff issue for the `fromisoformat` refactoring in the future (the more standard library the better IMHO, we need only LXML and request-like lib so far, so no new dependancy), that would depend on actual drop of Py3.6 support in pyodata - I created placeholder issue 190 for that. In this PR, let's stay on what you got already.

filak-sap:
Good job! I like this PR. What's the blocker? Why can't we merge it?

phanak-sap:
> Good job! I like this PR. What's the blocker? Why can't we merge it?

- We agreed that timezone supports needs to be done  (for consistency) for Edm.DateTime type as well. So far it seems that @rettichschnidi want to do it in same branch/PR (has created two other already), but after the small fixes, it can be of course merged as is, and add support for timezones in Edm.DateTime in second PR.
- There are still small fixes yet to be done from this code review, but also, they could be in another, small PR.
-  I also consider this can be merged to master, I simply leaved it still open for day or two and wait for answer from @rettichschnidi, how it is the easiest for him to continue.

rettichschnidi:
I will update my PR with all the improvements we have discussed. Please do not yet merge.

rettichschnidi:
Just updated with (hopefully) all changes incorporated.

Warnings/Questions:
* EdmDateTimeTypTraits:
  * I am not sure if and when the checks for None are actually needed in [from_json()](https://github.com/SAP/python-pyodata/blob/bfa215f138d7e76737fb6d2e268a513e30a7ba84/pyodata/v2/model.py#L419) and [from_literal()](https://github.com/SAP/python-pyodata/blob/bfa215f138d7e76737fb6d2e268a513e30a7ba84/pyodata/v2/model.py#L449). As far as I can see, None is already checked for in the respective functions in VariableDeclaration.
  * [to_json()](https://github.com/SAP/python-pyodata/blob/bfa215f138d7e76737fb6d2e268a513e30a7ba84/pyodata/v2/model.py#L410) seems to accept strings for [compatibility reasons](https://github.com/SAP/python-pyodata/blob/bfa215f138d7e76737fb6d2e268a513e30a7ba84/tests/test_service_v2.py#L630) (which I do not understand) - would this be needed for this for Emd.DateTimeOffset too?
* I added a 2nd commit which implements timezone handling for Emd.DateTimeOffset. This is what I tried to describe in the crossed out text above. My reasons for doing it the way I did are stated in the commit message. I do realize that this is opinionated and am OK with creating a separate PR for it.

tests/test_model_v2.py
rettichschnidi: Done

pyodata/v2/model.py
```diff
@@ -373,6 +363,40 @@ def from_literal(self, value):
...
 
 
+def ms_since_epoch_to_datetime(value, tzinfo):
+    """Convert milliseconds since midnight 1.1.1970 to datetime"""
+    try:
+        # https://stackoverflow.com/questions/36179914/timestamp-out-of-range-for-platform-localtime-gmtime-function
+        return datetime.datetime(1970, 1, 1, tzinfo=tzinfo) + datetime.timedelta(milliseconds=int(value))
+    except (ValueError, OverflowError):
+        min_ticks = -62135596800000
+        max_ticks = 253402300799999
```
phanak-sap, line 373: constants are usually uppercase.

phanak-sap:
Looks very good. Thanks for all the work.

PR #185, opened 2021-12-24
chore:  Update .travis.yml - Add Python 3.10 and latest LXML

phanak-sap:
Update compatibility matrix with new LXML and Python releases

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,8 @@ jobs:
     - python: 3.10.1
       env: LXML_VERSION=4.0.0     
     - python: 3.10.1
+      env: LXML_VERSION=4.1.1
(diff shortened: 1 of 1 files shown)


phanak-sap:
Seems I was overly optimistic, Python 3.10 is from Oct 5, but still not working correctly on Travis.
https://travis-ci.community/t/add-python-3-10/12220/10

phanak-sap:
There is almost month, where generic python "3.10" is not resolved, see https://travis-ci.community/t/add-python-3-10/12220/10

I will try to use as workaround specific version "3.10.1" + bionic. 

There may be worth, with leaving python 3.6, to leave Xenial as base distro in favor of Bionic or even Focal.

PR #186, opened 2021-12-31
[WIP] Use binary representation for Edm.Binary type values

inxonic:
This addresses two purposes:

1. Edm.Binary values can be uses as key elements
2. Values are stored in their Python native representation

While (1) could be addressed by converting between base16 and base64 in the from/to_literal methods, (2) is more consistent with other types like numbers and dates, which are also stored in their Python native representation.

The downside of (2) is that it breaks backward compatibility for cases where user code already expects the base64 representation for Edm.Binary values. Making this configurable might be an option. A change to some test cases is still required for this variant.

Please indicate wheather you are open to accept this change and which variant you prefer.


phanak-sap:
Before deciding (most probably @filak-sap), could you please elaborate the use case, when current pyodata is making problem for you? I am especially a bit puzzled by the statement 1, "Edm.Binary values can be uses as key elements" - what do you mean by that? Wrong URL generation? Or key elements where/for what?

Could you please for example post a sample python code? Or directly open new "bug" issue and link this PR with something which is ideally reproducible locally?

inxonic:
So, yes, my actual problem is URL generation. The entity uses an Edm.Binary as its key. When I access an item (i.e. of a search result), the URL element is generated as `Entity('BASE64')`, though the server only accepts `Entity(binary'BASE16')`.

inxonic:
Link to issue as requested: #187

phanak-sap:
Hi @inxonic, 

please don't feel discouraged, thanks for your work. I have this PR comments several times and checked the odata spec. Correct me if I am wrong please. We can move in two directions from here, but before I enumerate,  I would start with following premises:

 - PR that breaks backward compatibility will not be accepted. We are not at the moment expecting doing release of pyodata version 2 (if that would be needed, it would for something big like incompatible changes after adding support for odata v2/v4). 
 - PR with new feature but without tests will not be accepted as well. I would recommend starting with simple test that currently fails and reproduces your problem, even that now it is clearer from the PR comments.
 - OData specification V2 for Edm.Binary in both JSON and Atom formats (don't change just the JSON encoding/decoding!) expects Base64 encoding only. I was not able to find anything stating that Edm.Binary should be in other (base16, base32) encodings, as in RFC 3548.  For example this [odata spec issue](https://issues.oasis-open.org/browse/ODATA-466) was about that Base16 was for a time wrongly used for URLs, and Base64 should be used consistently to other protocol versions.
   - https://www.odata.org/documentation/odata-version-2-0/json-format/
   - https://www.odata.org/documentation/odata-version-2-0/atom-format/ 
   - https://www.rfc-editor.org/rfc/rfc3548.txt
- The problem part is therefore IMHO in the _"the server only accepts Entity(binary'BASE16')."_ - as I understand it at the moment - the server simply does not comply with odata v2 spec. But I understand it may harder to change that, than on the client side.
    
 Where to move now:

1) If you consider your problem to be unique to your integration only, I would suggest applying the PR code in your project and modify the model.py in runtime (monkey patching).

2) If you think that pyodata should support other than Base64 encoding, for whatever reason (even that it seems to be outside of the odata spec), than it needs to be a new configuration option, that would not break pyodata for current users. I am thinking that good place to start adding support different encoding/decoding options should be adding new parameter for [Config class constructor](https://github.com/SAP/python-pyodata/blob/21a99b50c3b1718e5b83109d6b4d944d06b7d1e5/pyodata/v2/model.py#L101-L103) - that would not be backwards incompatible. But then again, should there be support for all possible ways between Base16, Base32 and Base64 as in RFC 3548? So if it is done, its done thoroughly, and not just for one case (Base16<->Base64)?

inxonic:
Hi @phanak-sap, thanks for your detailed answer.

I completely understand that you cannot merge breaking PRs and that the problem has to be addressed in a different way.

However I still come to the conclusion that the service is compliant to the specs.

The odata spec issue you've linked adresses V4 and one [comment](https://issues.oasis-open.org/browse/ODATA-466?focusedCommentId=34226&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-34226) remarks "In OData 3.0 binary data in payloads is encoded with Base64, and binary data in URLs and CSDL documents is Base16 (HEX) encoded.", which is probably true for the older V2 also.

Further the [V2 spec](https://www.odata.org/documentation/odata-version-2-0/overview/) says in ch. 6: "Table 2 summaries the set of primitive types supported as well as how each MUST be represented when used in an OData URI or HTTP header. [...] binary'[A-Fa-f0-9][A-Fa-f0-9]*'". Though it does't actually name the encoding, the pattern only fits to base16.

So to me it looks like a V2 compliant client needs to convert Edm.Binary from base64 to base16 in case a value is taken from the payload and used as an entity key within a URL.

phanak-sap:
This will be another long comment. I divided it into section based on priority to read, while its construction came obviously from totaly opposite order of thinking. :)


**PR Code comments:**


- Bug you found is related to URL building only. But changes on model.py in commit 6827c04 will IMHO affect all three scenarios of pyodata usage - url building, direct consumption of the payoload data (read the value) and then sending the same values in some subsequent odata POST/PUT/MERGE requests. You don't want to change the binary encoding for the latter two scenarios, but stay in Base64 encoding. Tests are passing because this (change of binary encoding and check against some expected result) is IMHO not covered at all.
- Base64-base16 conversion should not be the only one considered (can be the only one implemented in the PR but be more open for future changes). 
- The more I think about this, the more I am sure that the **correct place to fix this** bug should be URL building in  service.execute(), just before calling `urlencode` ...:  https://github.com/SAP/python-pyodata/blob/21a99b50c3b1718e5b83109d6b4d944d06b7d1e5/pyodata/v2/service.py#L314
- ...and as stated before, with binary encoding passed as new `Config` constructor parameter:
 https://github.com/SAP/python-pyodata/blob/21a99b50c3b1718e5b83109d6b4d944d06b7d1e5/pyodata/v2/model.py#L101-L103


**Odata Specification comments:**


This PR - in essence a simple change - goes from "down to rabbit hole" into "Curios Case of Software Archaeology". So at least for possible future reference  :)

Odata version 2 was [Microsoft Open Specification Promise (OSP) ](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-odata/2b686a1a-9e1f-456f-80ff-072a010fc278 ). Then, Odata v4 become OASIS and later ISO standard. Mostly same as odata v2, though not backward-compatible.


But in the end, they build on top of the following RFCs:

- RFC3986 - Uniform Resource Identifier (URI): Generic Syntax - https://www.rfc-editor.org/rfc/rfc3986.txt
- RFC5023 The Atom Publishing Protocol - https://www.rfc-editor.org/rfc/rfc5023.txt
- Hypertext Transfer Protocol -- HTTP/1.1  - https://www.rfc-editor.org/rfc/rfc2616.txt

Every one does not restrain us (obviously) to use just one, correct binary encoding, but they reference other RFCs. So lower and lower we ends in something like this:


- RFC4648 - The Base16, Base32, and Base64 Data Encodings - https://www.rfc-editor.org/rfc/rfc4648.txt
- RFC5234 - Augmented BNF for Syntax Specifications: ABNF - https://www.rfc-editor.org/rfc/rfc5234.txt


And then, there is odata-abnf, for Odata v4, which in respect to URL building should be IMHO same as in odata v2 - 
[primitiveLiteral](https://github.com/oasis-tcs/odata-abnf/blob/742331219ee8cd26b18150d14a57cecf92469c3e/abnf/odata-abnf-construction-rules.txt#L876) and [binary](https://github.com/oasis-tcs/odata-abnf/blob/742331219ee8cd26b18150d14a57cecf92469c3e/abnf/odata-abnf-construction-rules.txt#L939-L944).

But even they [do not differentiate binary in URL and body in their test cases](https://github.com/oasis-tcs/odata-abnf/blob/main/abnf/odata-abnf-testcases.yaml#L274-L323) (contrary to say DateTime), so its IMHO back to openness of the HTTP and ATOM protocol itself, where the URI space of each server is controlled by that server - and Odata protocol imposes no further constrains, if the server conform to rules of query string construction.  

So in the end, I understand this all as, that any BaseX binary encoding can be used in the URL building, that will be in the end percent-encoded.


**Additional Notes:**

- Just out of curiosity, is the service you are connecting to, publicly available? If not, with what it was built - Apache Olingo? Something different? 

- Sadly there is no "odata v2 / v4 service, covering all test cases in the respective specification" that could be integration tests pointed against, to check if pyodata adhers to odata specification in its entirety. So everything goes "bottom-up".

- Pyodata was primary created for usage inside SAP - so tested on what we have in hand. We have never encountered Edm.Binary as key in URL, we are more using GUIDs for that purpose; just consume something in Edm.Binary, e.g. received PDF, image.

inxonic:
Thanks again for your detailed answer and for bringing in clarification by splitting the topics into sections.
I'd like to answer all of them (and change the order again):

**Additional Notes:**

- The service actually is a SAP Gateway.
- Indeed using Edm.Binary for a GUID value while there is a Edm.Guid primitive might not be optimal (though I fear the responsible people won't change that).
- However a UI5 application works well with this service (actually I found out which URLs the service accepts by using the browser debug mode with this app).
- It's behind a firewall, i.e. not publicly available.

**Odata Specification comments:**

I see the specification is quite complex, especially if you take the historyical development and the participating parties into account. However I'm now even more sure about its interpretation:

- The odata-abnf is for V4, i.e. not necessarily vaild vor V2.
- The specs link to a few RFC. But this is only for reference and relevant is what is inside the spces.
- Like said above the V2 spec states in the [Overview](https://www.odata.org/documentation/odata-version-2-0/overview/) (6.): "Table 2 summaries the set of primitive types supported as well as how each MUST be represented when used in an OData URI or HTTP header. Primitive type representations in request and response payloads are defined in the [OData:Atom] and [OData:JSON] specifications."
    "`'[A-Fa-f0-9][A-Fa-f0-9]*'`" &rarr; this is base16.
- The Microsoft [[MS-ODATA]](https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-ODATA/%5bMS-ODATA%5d.pdf) spec (2.2.2.) states more or less the same in slightly different words: "When the value of a primitive type needs to be represented in a URI or HTTP header, the representation that is written uses the primitive type literal form representation that is defined in the following table. A primitive type representation in a request or a response payload is defined in format-specific sections of this document."
    "`hexDigPair = 2*HEXDIG[hexDigPair]`" &rarr; this is also base16.

So to me, it boils down to:
- URI and HTTP header: base16
- For payload one needs to look at the detailed and format-specific sections.
    (Where one will find base64 for both Atom and JSON.)

Opposite to this (but out of scope) V4 uses the incompatible base64url encoding in all cases.

**PR code comments:**

- Changing the representation on the Python side was merely motivated by the pragmatical idiom of "killing two birds with one stone". Admittedly this is not a good practice in terms of software changes, especially when mixing up problem fixing and non backward compatible interface changes.
- I'm totally okay to take one step back and forget this PR or see it as a quick local fix and solely look at the original issue in #187 (maybe this would be even the better place for the other two points).

inxonic:
> * The more I think about this, the more I am sure that the **correct place to fix this** bug should be URL building in  service.execute(), just before calling `urlencode` ...:  https://github.com/SAP/python-pyodata/blob/21a99b50c3b1718e5b83109d6b4d944d06b7d1e5/pyodata/v2/service.py#L314

I'm afraid this might not be the correct place as this handles the query parameters.
The issue however is the encoding of the entity key which is a URI path component. So I think it's:
https://github.com/SAP/python-pyodata/blob/21a99b50c3b1718e5b83109d6b4d944d06b7d1e5/pyodata/v2/service.py#L372
which in turn leads to the `to_literal` method:
https://github.com/SAP/python-pyodata/blob/21a99b50c3b1718e5b83109d6b4d944d06b7d1e5/pyodata/v2/service.py#L210

phanak-sap:
> > * The more I think about this, the more I am sure that the **correct place to fix this** bug should be URL building in  service.execute(), just before calling `urlencode` ...:  https://github.com/SAP/python-pyodata/blob/21a99b50c3b1718e5b83109d6b4d944d06b7d1e5/pyodata/v2/service.py#L314
> 
> I'm afraid this might not be the correct place as this handles the query parameters. The issue however is the encoding of the entity key which is a URI path component. So I think it's:


Yes, not exactly this line, but this method and before this line. I.E. the fix should imho be in the ODataHttpRequest.execute() method, instead of model.py.

This line should be more applicable to your exact scenario: https://github.com/SAP/python-pyodata/blob/21a99b50c3b1718e5b83109d6b4d944d06b7d1e5/pyodata/v2/service.py#L302

But correct encoding of Edm.Binary as parameters in URL should be covered as well, right?

phanak-sap:
"The odata-abnf is for V4, i.e. not necessarily vaild vor V2." - Yes, not exactly valid, but IMHO this is the part that is backward compatible. Anyway, does not steer us away from that this issue needs to be fixed. 

I was in fact more focused on the HTTP protocol section RFC, which were even more broad, with `Encoding considerations: only "7bit", "8bit", or "binary" are  permitted `. So not only( !)  base16 for URI and HTTP headers, but perhaps simple octet Base8 (not covered by base64 python package in the PR)  or the Base64 without any change, if the odata server choose to support it as well. 

I am simply not want to stick with hardcoded Base64->Base16 and be open for more possibilities in the future, that would remain backward-compatible. That's all

filak-sap:
I like this PR. Unfortunately, it's to complex to understand the problem from written word, so would it be possible to examples and tests showing what was wrong and how it works after this PR is merged? (IOW would be nice to have tests to see what was changed and examles in commit message -> result before and results after).
Also, keeping the backward compatibility would be nice. Otherwise, I am all for having features simplifying users life instead of insisting on formal correctness.

phanak-sap (changes requested):
I marked it as "Request changes" to summarize progress so far.

1. we agreed this should be fixed
2. we discuss where and how the fix should be applied - still strongly support URL generation in service.py, not model.py changes
3. there definitely needs to be tests for the new changes.

phanak-sap:
Excellent commit message. I am humbled and @filak-sap will be pleased.

Btw, since you were now deep in the Edm.Binary.. would you be so kind to find time to handle this stalled PR correctly - #145?  :-)

tests/test_model_v2.py
```diff
@@ -1462,4 +1471,4 @@ def test_invalid_xml(xml_builder_factory):
 
     with pytest.raises(PyODataParserError) as e_info:
         MetadataBuilder(xml).build()
-    assert str(e_info.value) == 'Metadata document syntax error'
```
phanak-sap, line 1465: perhaps unnecessary change from editor defaults

phanak-sap (approved):

tests/test_model_v2.py
```diff
@@ -1462,4 +1471,4 @@ def test_invalid_xml(xml_builder_factory):
...
\ No newline at end of file
+    assert str(e_info.value) == 'Metadata document syntax error'
```
phanak-sap: Perhaps unnecessary change - editor defaults?

phanak-sap (approved):

tests/test_model_v2.py
filak-sap: That's good. Keep the change. Editors not adding new line at the end of file are not considered OK by me.

filak-sap:
Very well done! Thank you.

phanak-sap:
LOL, we both missed forgotten update of changelog.md.. updated in commit 6a4ac48

inxonic:
Gladly done. Thanks for merging this quickly.

> Btw, since you were now deep in the Edm.Binary.. would you be so kind to find time to handle this stalled PR correctly - #145? :-)

I've had a look. So it's about a case where `len(data: str)` is less than `sys.getsizeof(data: str)`. If I get it right _binary_ in that case means non-ASCII or multibyte characters. That's especially not the case for Edm.Binary which is Base64 and therefore a perfect subset of ASCII.

It's hard to guess the real issue with only a patch and no problem description or debug data. Maybe @bartonip can give some background.

Though I think he's right on the fact that there's a bug. And it's already in the SO article the code has been borrowed from.

https://github.com/SAP/python-pyodata/blob/ec56995551779cb99ae0a5734a7ecc3449e5191d/pyodata/v2/service.py#L137

Maybe the simplest solution would be to call read without parameter, which is [documented](https://docs.python.org/3/library/io.html#io.BufferedIOBase.read) to read until EOF.

bartonip:
Thanks for the ping @inxonic. The issue with len vs getsizeof is really that the HTTP response itself at the lowest level is a byte stream that gets translated into a string by the requests library. My understanding of what happens is that in the case of say, image data that is sent as pure binary (not as a base64 string represented as binary) when you run len on that resultant byte string the count will be too small due to bytes like \x00 not actually counting as a character in a string. Using getsizeof ensures that the actual size of the bytes in memory is returned and the buffer is created which is properly able to contain the response data.

Issue #78, opened 2019-12-10
unable to fetch data from the next page (more than 1000 records)


asmaMdi:
I'm using pyodata to get data from odata service but  but I'm getting only the first 1000 records. help me how to get it.
TIA

filak-sap:
Are you sure the service returns more entities in browser? I am not aware of any limit.
You might need to use $skip and $top parameters.
`*.get_entities().skip(1000).*`
`*.get_entities().top(2000).*`

asmaMdi:
in the browser the service returns the first 1000 records only. and there is a link to the next page at the end. I will try  .get_entities().skip(1000).top(2000).*
Thank you for your quick reply

asmaMdi:
Thank you. It works

filak-sap:
@asmaMdi May I know what service have this behaviour? We have a similar problem in #101 - it would be cool if we can compile a list of services which have interna implicit limits on get entity set size.

rettichschnidi:
@filak-sap Curia Vista, the business database of the Swiss Parlament (https://ws.parlament.ch/odata.svc) exposes this behavior and limits its responses to 1000 for some entities (e.g. MemberCommittee, MemberCommitteeHistory, MemberCouncilHistory, MemberParty)

PR #188, opened 2022-01-02
service: Support for server-side pagination (partial listings)

rettichschnidi:
Reading the spec, using the __next field, not $top/$skip/__count seems
to be the pristine way of fetching an entity collection in full:

> In response payloads, representing Collections of Entries, if the
> server does not include an object for every Entry in the Collection of
> Entries identified by the request URI then the response represents a
> partial listings of the Collection. In this case, "__next" name/value
> pair is included to indicate the response represents a partial
> listing. The value of the name/value pair is a URI which identifies
> the next partial set of entities from the originally identified
> complete set.

However, I am not sure if this the right way to implement this functionality. Feedback greatly appreciated.

Main question:
- What are the reasons that this project went with $top/$skip (#78) until now?

In case the maintainers agree with the direction of this PR I will work on getting it out of the draft state:
- Adding some example usage/documentation
- Adding a changelog entry

### DIFF
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -238,6 +238,7 @@ class ODataHttpRequest:
         self._headers = headers or dict()
         self._logger = logging.getLogger(LOGGER_NAME)
         self._customs = {}  # string -> string hash
+        self._next_url = None
 
     @property
     def handler(self):
@@ -299,7 +300,10 @@ class ODataHttpRequest:
 
            Fetches HTTP response and returns processed result"""
 
-        url = urljoin(self._url, self.get_path())
+        if self._next_url:
+            url = self._next_url
+        else:
+            url = urljoin(self._url, self.get_path())
         # pylint: disable=assignment-from-none
         body = self.get_body()
 
@@ -616,6 +620,17 @@ class QueryRequest(ODataHttpRequest):
             self._count = True
         return self
 
+    def next_url(self, next_url):
+        """
+        Sets URL which identifies the next partial set of entities from the originally identified complete set. Once
+        set, this URL takes precedence over all query parameters.
+
+        For details, see section "6. Representing Collections of Entries" on
+        https://www.odata.org/documentation/odata-version-2-0/json-format/
+        """
+        self._next_url = next_url
+        return self
+
     def expand(self, expand):
         """Sets the expand expressions."""
         self._expand = expand
@@ -667,6 +682,9 @@ class QueryRequest(ODataHttpRequest):
         }
 
     def get_query_params(self):
+        if self._next_url:
+            return {}
+
         qparams = super(QueryRequest, self).get_query_params()
 
         if self._top is not None:
@@ -1250,11 +1268,24 @@ class GetEntitySetRequest(QueryRequest):
 
 
 class ListWithTotalCount(list):
-    """A list with the additional property total_count"""
+    """
+    A list with the additional property total_count and next_url.
 
-    def __init__(self, total_count):
+    If set, use next_url to fetch the next batch of entities.
+    """
+
+    def __init__(self, total_count, next_url):
         super(ListWithTotalCount, self).__init__()
         self._total_count = total_count
+        self._next_url = next_url
+
+    @property
+    def next_url(self):
+        """
+        URL which identifies the next partial set of entities from the originally identified complete set. None if no
+        entities remaining.
+        """
+        return self._next_url
 
     @property
     def total_count(self):
@@ -1390,7 +1421,8 @@ class EntitySetProxy:
         return EntityGetRequest(get_entity_handler, entity_key, self)
 
     def get_entities(self):
-        """Get all entities"""
+        """Get some, potentially all entities"""
+
         def get_entities_handler(response):
             """Gets entity set from HTTP Response"""
 
@@ -1405,15 +1437,18 @@ class EntitySetProxy:
 
             entities = content['d']
             total_count = None
+            next_url = None
 
             if isinstance(entities, dict):
                 if '__count' in entities:
                     total_count = int(entities['__count'])
+                if '__next' in entities:
+                    next_url = entities['__next']
                 entities = entities['results']
 
             self._logger.info('Fetched %d entities', len(entities))
 
-            result = ListWithTotalCount(total_count)
+            result = ListWithTotalCount(total_count, next_url)
             for props in entities:
                 entity = EntityProxy(self._service, self._entity_set, self._entity_set.entity_type, props)
                 result.append(entity)
--- a/tests/test_service_v2.py
+++ b/tests/test_service_v2.py
@@ -2006,6 +2006,64 @@ def test_count_with_chainable_filter(service):
 
 
 @responses.activate
+def test_partial_listing(service):
+    """Using __next URI to fetch all entities in a collection"""
+
+    # pylint: disable=redefined-outer-name
+
+    responses.add(
+        responses.GET,
+        f"{service.url}/Employees?$inlinecount=allpages",
+        json={'d': {
+            '__count': 3,
+            '__next': f"{service.url}/Employees?$inlinecount=allpages&$skiptoken='opaque'",
+            'results': [
+                {
+                    'ID': 21,
+                    'NameFirst': 'George',
+                    'NameLast': 'Doe'
+                },{
+                    'ID': 22,
+                    'NameFirst': 'John',
+                    'NameLast': 'Doe'
+                }
+            ]
+        }},
+        status=200)
+
+    responses.add(
+        responses.GET,
+        f"{service.url}/Employees?$inlinecount=allpages&$skiptoken='opaque'",
+        json={'d': {
+            '__count': 3,
+            'results': [
+                {
+                    'ID': 23,
+                    'NameFirst': 'Rob',
+                    'NameLast': 'Ickes'
+                }
+            ]
+        }},
+        status=200)
+
+    # Fetching (potentially) all entities, actually getting 2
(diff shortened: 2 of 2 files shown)


phanak-sap:
Hi @rettichschnidi - you are very productive today with questions and PRs. :) Thanks for that, frankly we need that from userbase.

You must understand the history of this package. Sadly there is no "odata v2 / v4 service, covering all test cases in the respective specification" that could be integration tests pointed against, to check if pyodata adhers to odata specification in its entirety (feature coverage). So everything goes "bottom-up". Pyodata was primary created for usage inside SAP - for integration testing of Fiori Apps, because no python package was usable at that time. Then open-sourced.  

I guess, there is no big design decision for _"why this project went with $top/$skip",  more likely it is just` __next` is not implemented at all, since we covered our use cases with top and skip so far :)

pyodata/v2/service.py
```diff
@@ -616,6 +620,11 @@ def count(self, inline=False):
             self._count = True
         return self
 
+    def next_url(self, next_url):
+        """Sets the URL to be fetched. Takes precedence over all query parameters."""
```
phanak-sap, line 624: Link to odata spec pls

Vinuthan: Hi @phanak-sap,

This is also the same issue what we see for SAP SuccessFactors.

Sharing the documentation for future -
https://help.sap.com/viewer/d599f15995d348a1b45ba5603e2aba9b/2111/en-US/93ef8631b93b4d58be235b047dae2b57.html

pyodata/v2/service.py
```diff
@@ -1240,12 +1252,25 @@ def filter(self, *args, **kwargs):
         return self
 
 
-class ListWithTotalCount(list):
-    """A list with the additional property total_count"""
+class ListWithProgressIndication(list):
+    """
```
filak-sap, line 1256: Are you sure you need to rename the class?

filak-sap:
Impressive. I like this draft PR. Please, move it to "ready" PR.

> What are the reasons that this project went with $top/$skip (unable to fetch data from the next page (more than 1000 records) #78) until now?

Just because I am a lame OData user who knows only the bare minimum to be able to accomplish my job :)

pyodata/v2/service.py
rettichschnidi: I did it because it is more descriptive now. Not sure if it outweighs the API breakage however.

pyodata/v2/service.py
rettichschnidi: @Vinuthan I like the wording in the referenced documentation "Client-Side Pagination" vs "Server-Side Pagination"

rettichschnidi:
> My question - did you considered Batch requests/response?

I did not. My use case it "give me the whole database". I do not think batch requests can help me with that?

rettichschnidi:
> Pyodata was primary created for usage inside SAP - for integration testing of Fiori Apps, because no python package was usable at that time. Then open-sourced, "as-is".

I am glad you did this. Thanks!

docs/usage/querying.rst
```diff
@@ -31,8 +31,16 @@ Print unique identification (Id) and last name of all employees:
...
+    while True:
+        for employee in employees:
+            print(employee.EmployeeID, employee.LastName)
+
+        # Stop if server has no more entities left
+        if employees.next_url is None:
+            break
+
+        # We got a partial answer - continue with next page
+        employees = northwind.entity_sets.Employees.get_entities().next_url(employees.next_url).execute()
```
filak-sap, line 43: Oops, I didn't realize users will have to do this kind of exercise. Could you please create some Iterator wrapper? My goal was to create a library that integrates into python standard.

phanak-sap: @filak-sap I understand this does not look great. 

However, I would see also benefits in this approach. IMHO there is more control over the URLs sent, how it will look like (additional parameters) and how many requests are sent, when exactly and how big of response we receive. If hide subsequent requests behind iterator, that would yield entities based on presence of _next attribute, wouldn't we in the end transfer the entire DB to the memory in the model.py (for that particular entityset), regardless yielding of entity after entity out of the pyodata library to the client script? It is gut-feeling concern, I am not able to point to a code line that would do that.

rettichschnidi: > However, I would see also benefits in this approach. IMHO there is more control over the URLs sent, how it will look like (possible additional parameters) and how many requests are sent, when exactly and how big of response we receive.

+1 - We definitely should give the user control over this.

> If hide subsequent requests behind iterator, that would yield entities based on presence of _next attribute, wouldn't we in the end transfer the entire DB to the memory in the model.py (for that particular entityset), regardless yielding of entity after entity out of the pyodata library to the client script?

Returning a generator which yields individual entities (instead of a single `ListWithTotalCount`) could easily be implemented to not issue a 2nd request towards the OData server until all available entities have been consumed. This keeps the memory consumption on the level we have currently.

To avoid API breakage, how about this: Add a method `recursive()` to class `QueryRequest`. It could take a parameter `strategy` which controls whether to return a list of class `ListWithTotalCount` or a generator which either returns a generator of `ListWithTotalCount` or entity objects. Default would be the "strategy" which is currently implemented: Leaving it up to the user.

filak-sap: @rettichschnidi  Thank you for the proposal! Why not to make the comfortable strategy the default when you will add another method anyways? Could you please post some pseudo code?

rettichschnidi: I'd have to either break the API (switch to a generator) or keep all elements in memory (so that I can assemble a ListWithTotalCount which contains truly all elements) and therefore risking some kind of resource exhaustion.

Tell me which you you prefer and I will post some code.

jfilak: OK, the task seems to be too complex, so let's merge the current version and perhaps add the helper in future.

rettichschnidi: How about passing ListWithTotalCount? Something like this:

```python
employees = northwind.entity_sets.Employees.get_entities().next_page(employees).execute()
```

phanak-sap: IMHO throwing exception to in essence valid scenario is even worse than` if employees.next_url is None:` in a loop. Especially if the exception is related to odata protocol details.

**Option experimental**
We can take this in a way as an experimental API, (like they exists in Python 3.10.0 itself) - merge, not document, not in changelog, with follow-up issue that will either stick to it (and add documentation and release notes), change the API to some proposed in comments, or rollback the change.

phanak-sap:
> My question - did you considered Batch requests/response?
> I did not. My use case it "give me the whole database". I do not think batch requests can help me with that?

I was not exact enough. I ment consider in the solution - whatever it will be - several responses with _next field in one batch response. Should not be problem in current implementation IMHO, but perhaps with the iterator approach.

E.g. modified this test:
https://github.com/SAP/python-pyodata/blob/ec56995551779cb99ae0a5734a7ecc3449e5191d/tests/test_service_v2.py#L1338

phanak-sap:
The more I look at the docs/usage/querying.rst update, the more I must agree with @jfilak that the example of _next inside loop with `if employees.next_url is None:` somewhat strange to whatever else we've got in pyodata. 

I simply waited for his re-review, but I am inclined to move a bit forward, by leaving this out of upcoming 1.8.0 version - which I will do soon for what we already have stack up pending release. Better to be really sure how the new API should look like. I have marked this PR to be part of next MINOR 1.9.0 version.

docs/usage/querying.rst
rettichschnidi: > We can take this in a way as an experimental API, (like they exists in Python 3.10.0 itself) - merge, not document, not in changelog, with follow-up issue that will either stick to it (and add documentation and release notes), change the API to some proposed in comments, or rollback the change.

I could roll with this. @phanak-sap: Any chance you'd accept this for version 1.8?

rettichschnidi: I just pushed a new version without change note and no documentation change.

rettichschnidi:
> I was not exact enough. I ment consider in the solution - whatever it will be - several responses with _next field in one batch response. Should not be problem in current implementation IMHO, but perhaps with the iterator approach.

Now I get it - yes, certainly something to keep in mind, whatever the solution will look like.

docs/usage/querying.rst
phanak-sap: side note: well, they can always checkout the current code from your fork for this PR to play with as I did. Does not need to be merged to master for that. And by the force-push 

Where this PR will be merged as currently is, or not, there are tasks outside this PR to track. I created issue #198 for that, so far just to link, will add task-list after this PR is resolved. 

@jfilak What is your preference here? Merge current code to master? Request Changes? This is PR that I am not comfortable to approve or disapprove just on my own.

filak-sap:

rettichschnidi:
> Could you please rebase to the latest master (I don't like merge commits).

I did so. Anything else I can do to get this PR forward?

phanak-sap:
Hi @rettichschnidi  - basically I am waiting for answer from @jfilak / @filak-sap on my comment, how to proceed forward: https://github.com/SAP/python-pyodata/pull/188#discussion_r793477084.

Because of the `while true: if employees.next_url is None` discussion, I do not want currently to merge it to master just by myself - even undocumented, it would be de-facto pyodata API for foreseeable future. It can be played  locally even without  being on pypi, just by cloning your fork (and for your project you probably are already monkey patching). 

Note, it is definitely a valid enhancement request, it is tracked under #198 even if this PR would somehow ended closed without merging. I opted for being undecided on this PR for some more time.

filak-sap:
Well done. Thank you very much!

Issue #79, opened 2019-12-18
differentiate empty date


trancyQ:
Hi, 
Recently i am using pyodata to retrieve backend data.

Issue is that i still receive a datetime object (2000,01,01) for a date that is empty at the backend. And i saw this is the default date set during initialization.
Is there a way to tell whether the date i received is empty or it's a real date that is 2000 jan 1st.

Thanks.

jfilak:
Unfortunately, I cannot see any way how to differentiate between no date and 2000-01-01. Can you post the XML/JSON response from the server, please?

trancyQ:
i am not sure that make sense to do additional post requests since we are already getting data though this pyodata. 
Will it be possible a quick fix to change the default date to a more unrealistic date like 1000-01-01 or 1900-01-01? so that anyone look at it will know it's not an actual date. just suggestion.

trancyQ:
if no change will happen from your side, i am planning to do this.
if the date do not has a timezone information (.tzinfo) and the date is 2000-01-01, i will consider it as a default setting. Do you think this makes sense?

jfilak:
I would go for the lowest allowed value which is:

```
Edm.DateTime
Represents date and time with values ranging from 12:00:00 midnight, January 1, 1753 A.D.
```

https://www.odata.org/documentation/odata-version-2-0/overview/

I have no idea why we used 2000-01-01 - perhaps, that's how Business gateway handles Null DateTime values.

I asked for the response from the server to see how the server handles non-set date. If I learn how server handles it, I might be able to use `None` or something more appropriate.

jfilak:
I choose the value without thinking I guess and it make sense to change it to something less common.

trancyQ:
Thank you. this will be very helpful.

jfilak:
Sure, but can you please post your server response? Or at least the part where I can see what the server returns for empty date?

filak-sap:
@trancyQ  Would it work for you to use None for empty Date. I think it is a rational value.

rettichschnidi:
Once #189 got merged, users of this library will be able to detect missing properties and act on it - including setting a different/custom (default) time.

rettichschnidi:
https://github.com/SAP/python-pyodata/pull/189 got merged. Proposing to close this issue.

Issue #143, opened 2021-02-08
Date out of range error


sti0:
According to #80 we facing the same error. We're just consumer using the api to receive data for our DWH. Currently one of our job fails because of this error.

Is it possible to implement a fix for this problem on pyodata side?

Json:
```
{
    "d": {
        "results": {
            "__metadata": {
                "uri": "https://service.url/sap/c4c/odata/v1/c4codataapi/CorporateAccountIdentificationCollection('12345678910111213ABCDEFGHIJKLMNO')",
                "type": "c4codata.CorporateAccountIdentification",
                "etag": "W/"datetimeoffset'2021-02-01T19%3A41%3A17.9430080Z'""
            },
            "ObjectID": "12345678910111213ABCDEFGHIJKLMNO",
            "ParentObjectID": "166345678910111213ABCDEFGHIJKLMNO",
            "ETag": "/Date(1612208477943)/",
            "AccountID": "12345",
            "IDTypeCode": "ABCDE",
            "IDTypeCodeText": "<ABCD12>",
            "IDNumber": "1234-123",
            "ResponsibleInstitution": "",
            "EntryDate": null,
            "ValidFrom": "/Date(-62135769600000)/",
            "ValidTo": "/Date(253402214400000)/",
            "CountryCode": "",
            "CountryCodeText": "",
            "StateCode": "",
            "StateCodeText": "",
            "ZZWECK_KUT": "",
            "ZZWECK_KUTText": "",
            "CorporateAccount": {
                "__deferred": {
                    "uri": "https://service.url/sap/c4c/odata/v1/c4codataapi/CorporateAccountIdentificationCollection('12345678910111213ABCDEFGHIJKLMNO')/CorporateAccount"
                }
            }
        }
    }
}
```

Thanks & BR

filak-sap:
Well, we can for sure add a workaround (a global variable or something like that) which would cause that any value lower than -62135769600000 would be resolved as 1970-01-01T00:00:00. We haven't added it yet as one of the goals of PyOdata is to test OData service implementation and the value in your case is clearly wrong.

Feel free to open a PR. I would create a global variable in https://github.com/SAP/python-pyodata/blob/master/pyodata/v2/model.py which would enable a new test in at
https://github.com/SAP/python-pyodata/blob/6a4eb3b1b2ea29d3993ace76899f5bfbcb57ddc7/pyodata/v2/model.py#L416

```python
value_microseconds = int(value)
[...]

PR #189, opened 2022-01-02
service: Support null (#79)

rettichschnidi:
Until now, null values received from a OData server get substituted by a
default value.
Users of this library can not tell apart a value which coincidentally
equals the default, and a missing (null-ed) value received from the
server.

This commit allows the user to disable the substitution of null values by
default values, retrieving a None object instead.

Questions:
- Is this an acceptable approach to implement this features?

<s>In case the maintainers agree with the direction, I will add tests and documentation.</s> (done)

### DIFF
--- a/pyodata/client.py
+++ b/pyodata/client.py
@@ -69,7 +69,7 @@ class Client:
 
             # create service instance based on model we have
             logger.info('Creating OData Service (version: %d)', odata_version)
-            service = pyodata.v2.service.Service(url, schema, connection)
+            service = pyodata.v2.service.Service(url, schema, connection, config=config)
 
             return service
 
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -734,7 +734,7 @@ class EntityProxy:
        named values), and links (references to other entities).
     """
 
-    # pylint: disable=too-many-branches,too-many-nested-blocks
+    # pylint: disable=too-many-branches,too-many-nested-blocks,too-many-statements
 
     def __init__(self, service, entity_set, entity_type, proprties=None, entity_key=None, etag=None):
         self._logger = logging.getLogger(LOGGER_NAME)
@@ -761,11 +761,20 @@ class EntityProxy:
             # first, cache values of direct properties
             for type_proprty in self._entity_type.proprties():
                 if type_proprty.name in proprties:
+                    # Property value available
                     if proprties[type_proprty.name] is not None:
                         self._cache[type_proprty.name] = type_proprty.from_json(proprties[type_proprty.name])
-                    else:
+                        continue
+                    # Property value missing and user wants a type specific default value filled in
+                    if not self._service.retain_null:
                         # null value is in literal form for now, convert it to python representation
                         self._cache[type_proprty.name] = type_proprty.from_literal(type_proprty.typ.null_value)
+                        continue
+                    # Property is nullable - save it as such
+                    if type_proprty.nullable:
+                        self._cache[type_proprty.name] = None
+                        continue
+                    raise PyODataException(f'Value of non-nullable Property {type_proprty.name} is null')
 
             # then, assign all navigation properties
             for prop in self._entity_type.nav_proprties:
@@ -1581,10 +1590,11 @@ class FunctionContainer:
 class Service:
     """OData service"""
 
-    def __init__(self, url, schema, connection):
+    def __init__(self, url, schema, connection, config=None):
         self._url = url
         self._schema = schema
         self._connection = connection
+        self._retain_null = config.retain_null if config else False
         self._entity_container = EntityContainer(self)
         self._function_container = FunctionContainer(self)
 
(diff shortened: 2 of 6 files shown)


phanak-sap:
I think it is worth doing if there is genuine need for this - but not like additional hardcoded parameter for client and service classes, but instead new, optional parameter for the Config, that is already consumed by them. Default should be false, correct, to retain current behavior.

https://github.com/SAP/python-pyodata/blob/21a99b50c3b1718e5b83109d6b4d944d06b7d1e5/pyodata/v2/model.py#L101

It is the generic issue of specific problem of "Date valid minimum value vs default value" discussed in #143

But additionally, there is another layer of if-else possibilities - it depends also if the service metadata defines the attribute to be Nullable or not - see #102. if the `retain_null=True` and the service somehow returns Null on property that has metadata Nullable = False - it should fail obviously as currently is, not pass None object back.

rettichschnidi:
> I think it is worth doing if there is genuine need for this - but not like additional hardcoded parameter for client and service classes, but instead new, optional parameter for the Config, that is already consumed by them. Default should be false, correct, to retain current behavior.

I changed the Client to take the (extend) Config object. However, the server does not consume a Config, but a lit of arguments instead. Changing this would break the API.

> But additionally, there is another layer of if-else possibilities - it depends also if the service metadata defines the attribute to be Nullable or not - see #102. if the `retain_null=True` and the service somehow returns Null on property that has metadata Nullable = False - it should fail obviously as currently is, not pass None object back.

Fixed it in the latest push.

docs/usage/advanced.rst
```diff
@@ -105,3 +105,14 @@ If you need to work with many Entity Sets the same way or if you just need to pi
...
+Avoid default values for nulled values
+--------------------------------------
+
+To avoid missing property values getting substituted by their (type specific) default values, the service can be
+configured to retain missing values.
+
+.. code-block:: python
+
+    count = getattr(northwind.entity_sets, 'Employees').get_entities().count().execute()
+    print(count)
```
filak-sap, line 118: The documentation looks good except I am not sure how does the code sample relates to it?

filak-sap:
@rettichschnidi  How does this change solve #79? Does the reported need to configure the service instance with `retain_null=True`?

rettichschnidi:
> How does this change solve #79? 

Empty dates will return Null/None, not the default value.

> Does the reporter need to configure the service instance with `retain_null=True`?

Yes

docs/usage/advanced.rst
rettichschnidi: Not at all :face_with_spiral_eyes: - this slipped in by accident and the documentation part is still WIP. Would like to give a "real-world" example using Northwind, just need to find some entity which actually returns a null property.

rettichschnidi: Fixed

pyodata/v2/service.py
```diff
@@ -1581,10 +1590,11 @@ def function_import_handler(fimport, response):
 class Service:
     """OData service"""
 
-    def __init__(self, url, schema, connection):
+    def __init__(self, url, schema, connection, retain_null=False):
```
phanak-sap, line 1593: I do not particularly like this API. generic parameters, followed by boolean flag.. it is not much future-proof, when need arise for fifth or sixth configuration parameter in service. 

I understand that this is getting quite entangled an messy, because Config class is in the model.py module and so far was not used by service.py at at. And it should in fact be in the root next to client or exceptions, definitely above pyodata/v2, simply because same config parameters should work IMHO in future v4 support as well. Perhaps such move will be in the 2.0.0 version, where backward compatibility can be broken.. 

But service.py is already at the moment depending on model.py: https://github.com/SAP/python-pyodata/blob/0599db299cf5823f810b7806c2a8dda1227dde6e/pyodata/v2/service.py#L20

So why not add the `config=None` to Service, instead of  `retain_null=False`.. and in client.py:72 just pass the same value as in client.py:68

https://github.com/SAP/python-pyodata/blob/0599db299cf5823f810b7806c2a8dda1227dde6e/pyodata/client.py#L68

filak-sap:

rettichschnidi:
Rebased, improved commit messages and adjusted the documented example to #197.

pyodata/client.py
```diff
@@ -69,7 +69,7 @@ def __new__(cls, url, connection, odata_version=ODATA_VERSION_2, namespaces=None
 
             # create service instance based on model we have
             logger.info('Creating OData Service (version: %d)', odata_version)
-            service = pyodata.v2.service.Service(url, schema, connection)
+            service = pyodata.v2.service.Service(url, schema, connection, config=config)
```
phanak-sap, line 72: yeah, much better :)

phanak-sap (approved):
Looks good to me. Please resolve the conflict in changelog.md, so we have nice one commit as so far and we are ready to merge to master (and release 1.9.0)

Issue #190, opened 2022-01-04
chore: Drop support for Python 3.6, check compatibility with Python 3.10


phanak-sap:
Python 3.6 reached EOL (no security fixes) in 23 Dec 2021 and there is no actual need to stay backward-compatible. The lowest python version or this package should now be 3.7.

- [x] Add Python 3.10 to build matrix - #185
- [x] Drop Python 3.6 from travis builds - #190
- [x] Update the setup.py ( pipy documentation), update readme - #190
- [x] consider lowest LXML version possible to support based on 3.6 drop, update travis and setup.py dependencies accordingly - #190

phanak-sap:
all done, in 1.8.0

PR #192, opened 2022-01-11
Drop support for Python 3.6 

phanak-sap:
Part of issue #190


phanak-sap:
Should be Squashed to master.

PR #196, opened 2022-01-26
fix: more robust handling of XML EntityType Property with empty attributes

sh0rtcircuit:
If the `edmx` Schema of the Database contains empty attributes / empty strings, see updated `metadata.xml:339`, e,g.,
`<Property Name="Material_Type" Type="Edm.String" sap:text="" />`,

then `pyodata` throws an error:
`RuntimeError: The attribute sap:text of StructTypeProperty(Material_Type) is set to non existing Property `.

This fix accounts for empty attribute values, together with unit tests to test the behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Fix Increased robustness when schema with empty properties is returned
 
 ## [1.8.0]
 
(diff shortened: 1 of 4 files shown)


phanak-sap:
note: not sure why, but build was not triggered for this PR

sh0rtcircuit:
Any idea how I can retrigger it?

phanak-sap:
> Any idea how I can retrigger it?

Travis was now switched to Github actions. So far it seems no feature for manual triggering via some comment exists. Any new commit (-- amend + push ----force-with-lease, for example the rebasing requested by @filak-sap) should IMHO do the trick.

sh0rtcircuit:
> Could you please rebase to the latest master (I don't like merge commits).

Hi @filak-sap thanks for the comment, I updated it accordingly.

filak-sap:
Thank you very much for you contribution!

Issue #195, opened 2022-01-26
Error if Edm.DateTimeOffset is not set


metaodi:
This issue is based on an issue I got on a library that uses pyodata (https://github.com/metaodi/swissparlpy/issues/17)

Some calls fail when pyodata tries to parse dates:

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/opt/tljh/user/lib/python3.7/site-packages/pyodata/v2/model.py in parse_datetime_literal(value)
    388     try:
--> 389         return datetime.datetime.strptime(value, '%Y-%m-%dT%H:%M:%S.%f')
    390     except ValueError:

/opt/tljh/user/lib/python3.7/_strptime.py in _strptime_datetime(cls, data_string, format)
    576     format string."""
--> 577     tt, fraction, gmtoff_fraction = _strptime(data_string, format)
    578     tzname, gmtoff = tt[-2:]

/opt/tljh/user/lib/python3.7/_strptime.py in _strptime(data_string, format)
    358         raise ValueError("time data %r does not match format %r" %
--> 359                          (data_string, format))
    360     if len(data_string) != found.end():

ValueError: time data '0000-00-00T00:00:00' does not match format '%Y-%m-%dT%H:%M:%S.%f'
```

And the trace is the following:

```python
/opt/tljh/user/lib/python3.7/site-packages/swissparlpy/client.py in __init__(self, entity_request, variables)
     59 class SwissParlResponse(object):
     60     def __init__(self, entity_request, variables):
---> 61         self.entities = entity_request.execute()
     62         self.count = self.entities.total_count
     63         self.variables = variables

/opt/tljh/user/lib/python3.7/site-packages/pyodata/v2/service.py in execute(self)
    326             self._logger.debug('  body: <cannot be decoded>')
    327 
--> 328         return self._handler(response)
    329 
    330     def custom(self, name, value):

/opt/tljh/user/lib/python3.7/site-packages/pyodata/v2/service.py in get_entities_handler(response)
   1407             result = ListWithTotalCount(total_count)
   1408             for props in entities:
-> 1409                 entity = EntityProxy(self._service, self._entity_set, self._entity_set.entity_type, props)
   1410                 result.append(entity)
   1411 

/opt/tljh/user/lib/python3.7/site-packages/pyodata/v2/service.py in __init__(self, service, entity_set, entity_type, proprties, entity_key, etag)
    766                     else:
    767                         # null value is in literal form for now, convert it to python representation
[...]

PR #197, opened 2022-01-26
model: fix default Edm.DateTimeOffset value (#195)

rettichschnidi:
Default value was broken in several ways:
- Month 0 does not exist
- Day 0 does not exist
- Year 0 can not be handled by Python datetime

The new default value is copied from Edm.DateTime for consistency reasons. However, choosing something less common like 1.1.1 or 1.1.1970 might be more sensible?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
 - Fix Increased robustness when schema with empty properties is returned
+- Use valid default value for Edm.DateTimeOffset - Reto Schneider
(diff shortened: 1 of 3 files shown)


phanak-sap:
Well, since the Edm.DateTime default value is also causing problems, as you find out in issue #79, why not change both Edm.DateTime and Edm.DateTimeOffset to the January 1, 1753 A.D, noted in this comment

https://github.com/SAP/python-pyodata/issues/79#issuecomment-567090746

phanak-sap:
note: not sure why, but build was not triggered for this PR

filak-sap:

phanak-sap:
Thanks for the fix. Good work.

PR #201, opened 2022-01-30
Fix emd typos

OmniTroid:
Corrected Emd typos to Edm in several comments and error messages


phanak-sap (approved):
Thanks for not only noticing, but fixing right away.

PR #204, opened 2022-02-14
docs: fix typo in querying.rst

lkurija12:


phanak-sap:
Thanks for not only noticing - I guess by local error - but also correcting the documentation.

PR #205, opened 2022-02-22
model: Extract property 'FixedLength'

rettichschnidi:
Could not find any documentation on the default value if missing, but it seems is quite obvious that false, as opposed to true, is the only reasonable choice.

Please note: I have not defaulted the new parameter `fixed_length` in the `StructTypeProperty` constructor, because I got the impression that this class is not really meant to be part of the public API. Please let me know if you think otherwise.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - Client: Accept content type application/atom+xml - OmniTroid
+- Model: Extract property 'FixedLength' - Reto Schneider
 
 ## [1.9.0]
 
--- a/pyodata/v2/model.py
+++ b/pyodata/v2/model.py
@@ -780,7 +780,7 @@ class Collection(Typ):
 class VariableDeclaration(Identifier):
     MAXIMUM_LENGTH = -1
 
-    def __init__(self, name, type_info, nullable, max_length, precision, scale):
+    def __init__(self, name, type_info, nullable, max_length, precision, scale, fixed_length=None):
         super(VariableDeclaration, self).__init__(name)
 
         self._type_info = type_info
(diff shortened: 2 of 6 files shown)


pyodata/v2/model.py
```diff
@@ -1933,7 +1939,10 @@ def from_etree(entity_type_property_node):
             sap_attribute_get_string(entity_type_property_node, 'text'),
             sap_attribute_get_bool(entity_type_property_node, 'visible', True),
             sap_attribute_get_string(entity_type_property_node, 'display-format'),
-            sap_attribute_get_string(entity_type_property_node, 'value-list'), )
+            sap_attribute_get_string(entity_type_property_node, 'value-list'),
+            # End of TODO, regular attributes below
```
phanak-sap, line 1943: The TODO comment is probably a leftover

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - Client: Accept content type application/atom+xml - OmniTroid
+- Extract property 'FixedLength' - Reto Schneider
```
phanak-sap, line 11: "Model: Extract..."

phanak-sap:
The StructTypeProperty question: I am under impression that it is in fact public API. StructType is odata Edm type and the StructTypeProperty is used directly for example in tests:

https://github.com/SAP/python-pyodata/blob/0b353deca41b9bc5037ae525a71eb11297dc7212/tests/test_model_v2.py#L79

pyodata/v2/model.py
rettichschnidi: Evidently not a great comment, but it was meant to clearly separate the coming line from the sap_attribute_get... lines above, which have been commented as "TODO, regular attributes below". Will think of something better.

rettichschnidi:
Just pushed the proposed changes.

Issue #208, opened 2022-03-07
feature: Add async networking libraries support 


Albo90:
TOP LEVEL FEATURE TASK LIST:

- [x] create stable feature branch feature-branch - [async-feature](https://github.com/SAP/python-pyodata/tree/async-feature)
- [x] support for async networking - aiohttp.Client - client.py, service.py #210
- [x] check httpx package as networking client; should have sync and async API as well. If yes, update tests to have integration testing of requests (sync), aiohttp (async) and httpx (sync, async) packages
- [x] cover httpx async tests as well
- [x] documentation - https://github.com/SAP/python-pyodata/pull/266
- [x] integration to master - https://github.com/SAP/python-pyodata/pull/225
- [x] release - in version 1.10.0

Hi, 
I used this library recently, it's fantastic very convenient but it doesn't support **async pattern**. 
I and my collegue modified the library to support aiohttp, this is the link: https://github.com/Albo90/python-pyodata/tree/async_python
If you are interested, we can collaborate to integrate it in your project.
We are going to add test and more documentation in a couple of days.  

Thanks

filak-sap:
Impressive. Unfortunately, since you copied all files, I cannot see the needed changes. Can you somehow summarize what needs to be changed?

Albo90:
> Impressive. Unfortunately, since you copied all files, I cannot see the needed changes. Can you somehow summarize what needs to be changed?

Hi filak-sap, 
we had to create a bit of redundancy for our modifications otherwise it's really difficult for us to integragate async.  I try to describe them grouping for file. 

**async_client**
1) Definition of _new_ method accord to async strategy.
2) Definition of _fetch_metadata with get implementation accord to async.

**AsyncService**
1) Modification of execute method into OdataHttpRequest, we use "async with" to create request and build response in aim to avoid excessive ridefionition of project structure. 
2) Modification of http_get method. 

If you want, we can make a reasoning to build other ways of integration

filak-sap:
Is there a way how to avoid huge code duplications?

Albo90:
> Is there a way how to avoid huge code duplications?

Yes, in my option there is possibility to avoid creation of files (async_client, asyncService) rebuilding the existent methods.

Albo90:
@filak-sap we published a new version without duplication. The version is in this branch:   https://github.com/Albo90/python-pyodata/tree/no_duplication_v1. The library has a little syntax changement when it is used with async (pay attention to method getattr in example).

[...]

PR #210, opened 2022-03-09
Add async networking libraries support

Albo90:
- Part of #208
- intended to be merged to feature-branch only at the moment


phanak-sap:
I did checkout your code and try to run the following easy one async request example, modified from the generic sample from the issue record:

```
import asyncio
import aiohttp
from pyodata.client import Client

async def main():
    SERVICE_URL = 'http://services.odata.org/V2/Northwind/Northwind.svc/'

    async with aiohttp.ClientSession() as client:
      northwind = await Client.build_async_client(SERVICE_URL , client)
      order = northwind.entity_sets.Order_Details.get_entity(OrderID=10248, ProductID=42).execute()
      print(order.Quantity)
      
async_entities = asyncio.run(main())
```

Got following error:

```
  File "..\pyodata\v2\service.py", line 354, in _call_handler
    self._logger.debug('  url: %s', response.url)
AttributeError: '_RequestContextManager' object has no attribute 'url'
```

which is from this PR code snippet

 ```
   def _call_handler(self, response):
        self._logger.debug('Received response')
        self._logger.debug('  url: %s', response.url)
```

**Comment by Albo90** [comment]
> 

Syntax is different. I posted an example into issue yesterday, and I explained like syntax is different from sync version. In my opinion the main change is the getattr use because we can't transform __getattribute__ as async without huge and maybe dangerous impact.

Your example with correct syntax:

```python

import asyncio

import aiohttp
from pyodata.client import Client


async def main():
    SERVICE_URL = 'http://services.odata.org/V2/Northwind/Northwind.svc/'

    async with aiohttp.ClientSession() as client:
        northwind = await Client.build_async_client(SERVICE_URL, client)
        order = await northwind.entity_sets.Order_Details.get_entity(OrderID=10248, ProductID=42).async_execute()
        print(await order.getattr('Quantity'))


async_entities = asyncio.run(main())
 ```

**Comment by phanak-sap** [comment]
Oh, I see my error, I forgot to change the `execute() `to `async_execute()`. Of course  - knew about it and still:)
Works OK.

File: pyodata/v2/response.py [hunk 1]
```diff
@@ -0,0 +1,32 @@
...
+Utility class to standardize response
+
+Author: Alberto Moio <email Alberto>, Nunzio Mauro <mnunzio90@gmail.com>
+Date:   2017-08-21
+"""
+import json
+
+
+class Response:
+    """Representation of http response in a standard form already used by handlers"""
```
phanak-sap, line 11: Odata v2 has both JSON and XML Atom as standard response...
https://www.odata.org/documentation/odata-version-2-0/atom-format/

pyodata/v2/service.py
```diff
@@ -858,6 +889,19 @@ def __getattr__(self, attr):
                 raise AttributeError('EntityType {0} does not have Property {1}: {2}'
                                      .format(self._entity_type.name, attr, str(ex)))
 
+    async def getattr(self, attr):
```
phanak-sap, line 892: I guess this method should have async_ prefix as well, similar to other new ones.

Will throw RuntimeError when called in synchronous script, where it looks like correct usage until seen the pyodata source code I guess.

```
import requests
import pyodata

import logging
logging.basicConfig(level=logging.DEBUG)

SERVICE_URL = 'http://services.odata.org/V2/Northwind/Northwind.svc/'

northwind = pyodata.Client(SERVICE_URL, requests.Session())
order = northwind.entity_sets.Order_Details.get_entity(OrderID=10248, ProductID=42).execute()
print(order.getattr('Quantity'))
```
```

<coroutine object EntityProxy.getattr at 0x000002032DFD7CA0>
...: RuntimeWarning: coroutine 'EntityProxy.getattr' was never awaited
  print(order.getattr('Quantity'))
```

phanak-sap:
@Albo90 Totally lame question (async python knowledge) - could you please provide sample script where the new async `.getattr` should be used as intended; compared to getting the value old vay trough non-async  `__getattr__`? I guess passing to another async function (not just logging or printing).

You can modify your example hardcoded against Nortwind, with both getters being used:

```
import asyncio
import aiohttp
from pyodata.client import Client

import logging
logging.basicConfig(level=logging.DEBUG)

import asyncio
async def main():
    async with aiohttp.ClientSession() as client:
        logger = logging.getLogger("SCRIPT")
        SERVICE_URL = 'http://services.odata.org/V2/Northwind/Northwind.svc/'
        service = await Client.build_async_client(SERVICE_URL, client)
        entities = await service.entity_sets.Orders.get_entities().async_execute()

        for entity in entities:
            attr_value = await entity.getattr("OrderID")            
            logger.info('Order ID value (version: %d)', attr_value)
            print(attr_value)
            print(entity.OrderID,entity.EmployeeID,entity.CustomerID)
        return entities

async_entities = asyncio.run(main())
```

Albo90:
> @Albo90 Totally lame question (async python knowledge) - could you please provide sample script where the new async `.getattr` should be used as intended; compared to getting the value old vay trough non-async `__getattr__`? I guess passing to another async function (not just logging or printing).
> 
> You can modify your example hardcoded against Nortwind, with both getters being used:
> 
> ```
> import asyncio
> import aiohttp
> from pyodata.client import Client
> 
> import logging
> logging.basicConfig(level=logging.DEBUG)
> 
> import asyncio
> async def main():
>     async with aiohttp.ClientSession() as client:
>         logger = logging.getLogger("SCRIPT")
>         SERVICE_URL = 'http://services.odata.org/V2/Northwind/Northwind.svc/'
>         service = await Client.build_async_client(SERVICE_URL, client)
>         entities = await service.entity_sets.Orders.get_entities().async_execute()
> 
>         for entity in entities:
>             attr_value = await entity.getattr("OrderID")            
>             logger.info('Order ID value (version: %d)', attr_value)
>             print(attr_value)
>             print(entity.OrderID,entity.EmployeeID,entity.CustomerID)
>         return entities
> 
> async_entities = asyncio.run(main())
> ```

I answer you about async and coroutine, tomorrow I send you more details for each point. For me the most important information about async is to understand that an async method cannot be used like sync otherwise the result is "RuntimeWarning: coroutine".  For that reason getattr is only for async and we didn't use your __getattribute__, it's a python standard method and it's sync, if we changed it with async (I think it's impossible but I use it to explain you) you standard code doesn't work .

phanak-sap:
Yeah I run into "RuntimeWarning: coroutine" when playing with your code and I understand why is that. I will be happy for any answer of course, but it is valid "just read PEP 492 and asyncio/aiohttp documentation" also :)

So far it seems the new stuff  from this PR is not affecting the old, so no backward incompatible changes for standard, synchronous usage of pyodata. That's is the main focus for the feature in fact, no breaking of existing code out there. Pyodata should remain working correctly with synchronous and standard multithreading usage. 

I asked basically for a simplest snippet possible for documentation, I know that little bit early on.

Albo90:
In the evening we will add these feature:

- Async prefix to method getattr.
- Atom format for the response
- Documentation and test (at least a part of those)

For the async argument, I attach an example:
```python
import asyncio

import aiohttp
from pyodata.client import Client


async def main():
    SERVICE_URL = 'http://services.odata.org/V2/Northwind/Northwind.svc/'

    async with aiohttp.ClientSession() as client:
        northwind = await Client.build_async_client(SERVICE_URL, client)
        orders = await northwind.entity_sets.Order_Details.get_entities().select('OrderID,ProductID').async_execute()
        for order in orders:
            # quantity = await order.getattr('Quantity')
            quantity = order.Quantity
            print(quantity)


async_entities = asyncio.run(main())
```
An async method like getattr need "await" when it is called. If you don't set it, you create only coroutine. The coroutine cannot be print or interrogate, it can be only executed with await. The problem is your implementantion of __ getattribute __ uses cache to permit for example "order.Quantity" when you use async it could have a problem so we defined getattr, if you want I can make an example where "order.Quantity" doesn't work but "await order.getattr("Quantity")" works.

phanak-sap:
Hi @Albo90
 I am sorry, we do not yet have enabled tests for this feature branch as well. At the moment they run against master PRs only. But all your new tests are failing (my laptop as a CI server :) ).

You are perhaps using libraries that are part of your system PIP installed packages, but missing that in the PR code.
(update dev-requirements.txt file accordingly)

Please use following workflow for your local work:

- use virtualenv, so you are isolated from your system-level installed packages
- activate the virtualenv 
- pip install -r dev-requirements.txt
- pip install -r requirements.txt
- pytest

All tests should pass before submitting to the PR.


Errors I see with the checkout of currently latest commit a4cc84d6a:

```
PytestUnhandledCoroutineWarning: async def functions are not natively supported and have been skipped.
  You need to install a suitable plugin for your async framework, for example:
    - anyio
    - pytest-asyncio
    - pytest-tornasync
    - pytest-trio
    - pytest-twisted
    warnings.warn(PytestUnhandledCoroutineWarning(msg.format(nodeid)))

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
===================================== 242 passed, 8 skipped, 14 warnings in 4.46s =====================================
```

Even if I install pytest-asyncio plugin, the tests are still failing:

```

________________________________ ERROR at teardown of test_client_custom_configuration ________________________________

    @pytest.fixture
    def event_loop():
        loop = asyncio.new_event_loop()
        yield loop
>       loop.run_until_complete(asyncio.sleep(0.1, loop=loop))
E       TypeError: sleep() got an unexpected keyword argument 'loop'
```

phanak-sap:
@Albo90  please agree to the CLA  - see https://github.com/SAP/python-pyodata/blob/master/CONTRIBUTING.md#developer-certificate-of-origin-dco

dev-requirements.txt
```diff
@@ -1,4 +1,6 @@
 requests==2.23.0
+pytest-asyncio == 0.15.1
+aiohttp==3.8.1
```
phanak-sap, line 3: I have already updated as first commit to the async-feature branch the dev-dependency.txt `pytest-aiohttp>=1.0.4`. 

Wouldn't explicit pytest plugin by better than generic aiohttp library? Just asking what dependency is better, doesn't seem right to have both.

Albo90: We don't have errors in our test, I forgot the requirements. We added our dev-requirements.txt. We don't have aiohttp because we want to stay generic without aiohttp dependency.

phanak-sap: I must say I am not understanding the last comment at all.

1. what does "We don't have aiohttp test because we want to stay generic without aiohttp dependency." mean? Especially when you commited  `aiohttp==3.8.1` to the dev-dependencies already? Do you want that or not?

2. And should the `pytest-aiohttp` from me, which already is in the `async-feature` branch dev-requirements.txt  stay or not? Should the async tests be based on pytest-aiohttp plugin or pytest-asyncio +aiohttp ? I expected we will go pytest plugins way.

tests/conftest.py
```diff
@@ -1,4 +1,5 @@
 """PyTest Fixtures"""
+import asyncio
```
phanak-sap, line 2: You are importing direct asyncio package, while in dev dependencies is pytest-asyncio plugin.

Therefore tests are still failing.

Btw, why you prefer the asyncio/aiohttp directly as a dependency and not the pytest plugins, if it is used for testing purposes only anyway? My guess would be that pytest plugin are used by many and doing IMHO the same stuff you are doing by hand - according to readme of both plugins. In this case I think is best to use standard solution of the problem instead of own implementation.

dev-requirements.txt
phanak-sap: Question: is your code in client.py/service.py compatible with `httpx` python package? 
Or does it expect `aiohttp.ClientSession()` to be provided in the `Client.build_async_client` only?

If it is more open (best possible outcome, I don't actually know about any other than `requests` library being used with pyodata, see for example https://github.com/SAP/python-pyodata/pull/154#issuecomment-863462533 ), then why not add both integration tests with `aiohttp `and `httpx `packages?

tests/test_async_client.py
```diff
@@ -0,0 +1,129 @@
+"""PyOData Client tests"""
+from unittest.mock import patch, AsyncMock
+
+import aiohttp
```
phanak-sap, line 4: These tests are now passing, so we are from 8 failing down to 6 .)

tests/test_async_client.py
```diff
@@ -0,0 +1,129 @@
...
+
+    app = web.Application()
+    app.router.add_get('/$metadata', generate_metadata_response(headers={'content-type': content_type}, body=metadata))
+    client = await aiohttp_client(app)
+
+    service_client = await Client.build_async_client(SERVICE_URL, client)
+
+    assert isinstance(service_client, pyodata.v2.service.Service)
+
+    # one more test for '/' terminated url
```
phanak-sap, line 57: mark at least as TODO to grab more attention in future PR if not done.

dev-requirements.txt
mnunzio: We should check, in theory it should work with httpx as well.

tests/conftest.py
mnunzio: We forgot to update dev-requirements.txt now you shouldn't have any problems with the test anymore. We have removed some useless code.

phanak-sap: Yeah, with 3bab585 I have passing tests, only with one warning. 

```
pytest_aiohttp\plugin.py:28: DeprecationWarning: The 'asyncio_mode' is 'legacy', switching to 'auto' for the sake of pytest-aiohttp backward compatibility. Please explicitly use 'asyncio_mode=strict' or 'asyncio_mode=auto' in pytest configuration file.
    config.issue_config_time_warning(LEGACY_MODE, stacklevel=2)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================================== 250 passed, 1 warning in 3.55s
```

PR #211, opened 2022-03-14
docs: fix a NameError in querying.rst

allan-cedric:
A simple NameError when executing 'Get entities matching a filter' code. Instead of 'smith_employees', it should be 'smith_employees_request'.


phanak-sap:
Thanks for noticing and fixing right away. Good job.

phanak-sap (approved):

PR #220, opened 2022-07-10
docs: cover URLs generation topic

phanak-sap:
+ add test that covers both custom() in a POST request and get_body() - not covered so far
+ add doc example of pyodata as url provider for other libraries

### DIFF
--- a/docs/usage/urls.rst
+++ b/docs/usage/urls.rst
@@ -1,0 +1,43 @@
+URLs generation
+===============
+
+.. _Locust: https://docs.locust.io/en/stable/
+
+Sometimes you may want to not use **PyOData** to actually make the HTTP requests, but
+just grab the url and body for some other library. For that, you can use following
+methods from ODataHttpRequest class - which is base of every query, update or delete
+covered in pyodata documentation.
+
+.. code-block:: python
+
+  .get_method()
+  .get_path()
+  .get_query_params()
+  .get_body()
+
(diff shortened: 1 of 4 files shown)


docs/usage/urls.rst
```diff
@@ -0,0 +1,43 @@
...
+  smith_employees_query = odataClient.entity_sets.Employees.get_entities().filter("FirstName eq 'John' and LastName eq 'Smith'")
+
+  class MyUser(HttpUser):
+      wait_time = between(5, 15)
+      host = SERVICE_URL
+
+      @task(1)
+      def filter_query(self):
+          urlpath = smith_employees_query.get_path()
+          url = SERVICE_URL + urlpath
```
mnezerka, line 41: Wouldn't it be better to use e.g. `os.path.join()` for this (to avoid issues with slashes)?

mnezerka (approved):
Looks good to me (just one very minor comment regarding doc example)

docs/usage/urls.rst
phanak-sap: I thought about that actually, the SERVICE_URL is hardcoded in my sample and get_path() seems to return url part always without leading '/' character. The sample script works as is - so I wanted to exclude what I thought is not necessary. 

But yeah, it makes it more "safe to copy-pasting" example, so I will update it with `os.path.join()`

PR #222, opened 2022-08-07
ci: update LXML package version list

phanak-sap:


phanak-sap:
all github actions passing OK

PR #223, opened 2022-08-08
service: fix navigation for associations with many to one relationships and for associations with relationships to same entity

Masterchen09:
When playing around with the [OData API of the SAP Analytics Cloud](https://api.sap.com/api/Story_API/resource) to retrieve some stories, I have noticed that the nav method did not work with many to one relationships. The SAC API specifies such an relationship between a story and users (createdByUser, a story can be created only by one user, but a user can create many stories): Either you have to specify a key for the nav method or to use the get_entities method which both not works as the target entity is not a list of entities (the entity is returned directly and not in a results array) or as an invalid url is generated (the target entity is being indexed, which does not work as it is not an list).

I have tracked down this issue down to line [886](https://github.com/SAP/python-pyodata/blob/08545ebcf9709642ebc7992af3e82c345869a1ff/pyodata/v2/service.py#L886) (and [1362](https://github.com/SAP/python-pyodata/blob/08545ebcf9709642ebc7992af3e82c345869a1ff/pyodata/v2/service.py#L1362)) where the multiplicity for all roles is checked and when the multiplicity for one of the role is "*" the check fails. Here only the multiplicity of the "ToRole" should be checked as this role is the relevant role for the navigation.

Additionally when writing a test for this case I have found another issue: It is not possible to use the nav method with associations where both ends are of the same entity type. The logic for finding the navigation_entity_set in line [877-880](https://github.com/SAP/python-pyodata/blob/08545ebcf9709642ebc7992af3e82c345869a1ff/pyodata/v2/service.py#L877-L880) (and [1352-1359](https://github.com/SAP/python-pyodata/blob/08545ebcf9709642ebc7992af3e82c345869a1ff/pyodata/v2/service.py#L1352-L1359)) is using the entity set name for finding the role in the association set: This does not work if both roles has the same entity set name. Here the method end_by_role method of the association set can be used to find the correct role (and then the correct navigation_entity_set).

### DIFF
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -874,16 +874,10 @@ class EntityProxy:
             association_info.name,
             association_info.namespace)
 
-        navigation_entity_set = None
-        for end in association_set.end_roles:
-            if association_set.end_by_entity_set(end.entity_set_name).role == navigation_property.to_role.role:
-                navigation_entity_set = self._service.schema.entity_set(end.entity_set_name, association_info.namespace)
+        end = association_set.end_by_role(navigation_property.to_role.role)
+        navigation_entity_set = self._service.schema.entity_set(end.entity_set_name)
 
-        if not navigation_entity_set:
-            raise PyODataException(f'No association set for role {navigation_property.to_role}')
-
-        roles = navigation_property.association.end_roles
-        if all((role.multiplicity != model.EndRole.MULTIPLICITY_ZERO_OR_MORE for role in roles)):
+        if navigation_property.to_role.multiplicity != model.EndRole.MULTIPLICITY_ZERO_OR_MORE:
             return NavEntityProxy(self, nav_property, navigation_entity_set.entity_type, {})
 
         return EntitySetProxy(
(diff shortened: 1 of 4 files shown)


phanak-sap:
Hi @ Masterchen09 
 Thank you for not only uncovering more than one problem, but also providing fix for all of them.

pyodata/v2/service.py
```diff
@@ -874,16 +874,10 @@ def nav(self, nav_property):
...
 
-        navigation_entity_set = None
-        for end in association_set.end_roles:
-            if association_set.end_by_entity_set(end.entity_set_name).role == navigation_property.to_role.role:
-                navigation_entity_set = self._service.schema.entity_set(end.entity_set_name, association_info.namespace)
+        end = association_set.end_by_role(navigation_property.to_role.role)
+        navigation_entity_set = self._service.schema.entity_set(end.entity_set_name)
 
-        if not navigation_entity_set:
-            raise PyODataException(f'No association set for role {navigation_property.to_role}')
```
phanak-sap, line 883: If the bug is about "many to one relationships", why remove this exception rise? Even the second part of the PR description - "it is not possible to use the nav method with associations where both ends are of the same entity type." - as I understand it, the case that you would like to refer to something that does not exist at all is still valid. 

As an example - consider metadata file, that is invalid, but [forced to be parsed anyway ](https://github.com/SAP/python-pyodata/blob/master/docs/usage/initialization.rst#dealing-with-errors-during-parsing-metadata)  - and then you can end easily with trying to access something which is not there at all.

It bothers me a bit that there is not a test that would start to fail by removing these lines - you are only adding new tests after all.

Masterchen09: I have removed this exception because the end_by_role method is [raising a KeyError](https://github.com/SAP/python-pyodata/blob/9a72f930800ed758cf8a4d9c53a8c2434ff49133/pyodata/v2/model.py#L2165) in case the end role was not found. I have thought about to handle the KeyError and raising the PyODataException, but this would be a bit redundant. However, due to the possible KeyError of end_by_role the navigation_entity_set can never be None (e.g. in case of an invalid metadata file) and the PyODataException would never be raised.

pyodata/v2/service.py
```diff
@@ -874,16 +874,10 @@ def nav(self, nav_property):
             association_info.name,
             association_info.namespace)
 
-        navigation_entity_set = None
-        for end in association_set.end_roles:
```
phanak-sap, line 878: why was the for loop removed? Again, all tests are still passing, but this may be a problem of the test suite - that the loop was not covered so far, just simple cases around navigational properties.

Masterchen09: The loop is not necessary (anymore) because the end_by_role method of the association_set is "looping" over the association_set.end_roles using the generator expression here: https://github.com/SAP/python-pyodata/blob/9a72f930800ed758cf8a4d9c53a8c2434ff49133/pyodata/v2/model.py#L2163

phanak-sap: OK, that's true. And also, since now test metadata.xml contains multimple association sets ending to same end role (Customer), it is proven by tests.

phanak-sap (approved):
All questions answered, all builds are passing, looking good.

Thanks for the good work.

PR #225, opened 2022-08-11
feature: async support

phanak-sap:
This is basically the code from PR #210 (that one merged to feature-branch), by @Albo90, with more tests.

This mess will be squashed to one "feature" commit to master.


pyodata/client.py
```diff
@@ -37,6 +53,25 @@ class Client:
...
+
+        logger = logging.getLogger('pyodata.client')
+
+        if odata_version == Client.ODATA_VERSION_2:
+
+            # sanitize url
+            url = url.rstrip('/') + '/'
+
+            if metadata is None:
+                metadata = await _async_fetch_metadata(connection, url, logger)
```
jfilak, line 69: Don't you kill all async here? What's the difference between the regular client?

phanak-sap: no, await (asynchronous wait) is calling the asynchronous _async_fetch_metadata function.

PR #235, opened 2022-10-03
pin pytest, pytest-cov, sphinx dev-requirements

phanak-sap:
In PR #234, all the python 3.7 builds (but only .37) failed on flake8 step of linting.

File "/opt/hostedtoolcache/Python/3.7.14/x64/lib/python3.7/site-packages/flake8/plugins/manager.py", line 254, in _load_entrypoint_plugins
    eps = importlib_metadata.entry_points().get(self.namespace, ())
AttributeError: 'EntryPoints' object has no attribute 'get'

Most probable cause is the recent release of pytest-cov-4.0.0  and upgrades/mismatch in its transitive dependencies.  Pinning so builds at least run in same config as week ago and failing CI is not blocking PRs

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,5 +1,5 @@
 requests>=2.28.1
-pytest>=7.1.2
+pytest==7.1.3
 pytest-aiohttp>=1.0.4
 responses>=0.21.0
 httpx>=0.23.0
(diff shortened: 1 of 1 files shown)


phanak-sap:
no longer applicable

PR #241, opened 2023-01-09
Sharepoint support

jenrik:
Add experimental support for Microsoft's odata variant that is used in Sharepoint

Minimal usage example with NTLM authentication
```python
import requests
from requests_ntlm3 import HttpNtlmAuth
import pyodata
from pyodata.vendor.sharepoint import (
    SPEdmDateTimeTypTraits
)
from pyodata.v2.model import (
    Types,
    Typ
)

Types.overwrite_type(Typ('Edm.DateTime', '1753-01-01T00:00', SPEdmDateTimeTypTraits()))

SERVICE_URL = 'https://example.com'

# Create instance of OData client
session = requests.Session()
session.auth = HttpNtlmAuth("<username>", "<password>")
client = pyodata.Client(SERVICE_URL, session)
```

Requirements for NTLM authentication
```
# Patched ntlm-auth that implements md4 because it has been removed from openssl/python
git+https://github.com/Sleipner-Motor-AS/ntlm-auth.git@520c21ac8d939bd6f4e62446dd0c4ac7ae0abf4a
requests-ntlm3
```

### DIFF
--- a/pyodata/v2/model.py
+++ b/pyodata/v2/model.py
@@ -235,6 +235,22 @@ class Types:
             Types.Types[collection_name] = collection_typ
 
     @staticmethod
+    def overwrite_type(typ):
+        """Add new  type to the type repository as well as its collection variant"""
+
+        # build types hierarchy on first use (lazy creation)
+        if Types.Types is None:
+            Types._build_types()
+
+        # pylint: disable=unsupported-assignment-operation
+        Types.Types[typ.name] = typ
+
+        collection_name = f'Collection({typ.name})'
+        collection_typ = Collection(typ.name, typ)
+        # pylint: disable=unsupported-assignment-operation
+        Types.Types[collection_name] = collection_typ
+
+    @staticmethod
     def from_name(name):
 
         # build types hierarchy on first use (lazy creation)
@@ -755,7 +771,7 @@ class Collection(Typ):
         self._item_type = item_type
 
     def __repr__(self):
-        return f'Collection({repr(self._item_type)})'
+        return f'Collection({self._item_type.name})'
 
     @property
     def is_collection(self):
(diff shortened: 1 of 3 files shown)


pyodata/v2/model.py
```diff
@@ -234,6 +234,22 @@ def register_type(typ):
             # pylint: disable=unsupported-assignment-operation
             Types.Types[collection_name] = collection_typ
 
+    @staticmethod
+    def overwrite_type(typ):
+        """Add new  type to the type repository as well as its collection variant"""
```
phanak-sap, line 239: This is probably copypasted docstring, does not connect to the method name.

This should definitely, as a new feature around Types, be covered with new unit tests, that it behave as expected.
- new type
- redefinition of existing type
- ...whatever I missed

pyodata/v2/model.py
```diff
@@ -234,6 +234,22 @@ def register_type(typ):
...
+
+        # build types hierarchy on first use (lazy creation)
+        if Types.Types is None:
+            Types._build_types()
+
+        # pylint: disable=unsupported-assignment-operation
+        Types.Types[typ.name] = typ
+
+        collection_name = f'Collection({typ.name})'
```
phanak-sap, line 248: Keep in mind this commit
https://github.com/SAP/python-pyodata/commit/bfa215f138d7e76737fb6d2e268a513e30a7ba84

phanak-sap:
Hi @jenrik,
 first of all, good work so far, I can imagine the problem at hand and the work behind this PR.

Basically polish the PR according to https://github.com/SAP/python-pyodata/blob/master/CONTRIBUTING.md#did-you-write-a-patch

- Please sign the  CLA, cannot be merged without it.
- fix existing linting and failing tests
- add new test for the new feature
- update the documentation.

question: does the example in PR description need this?
`Types.overwrite_type(Typ('Edm.DateTime', '1753-01-01T00:00', SPEdmDateTimeTypTraits()))`

Shouldn't that be in the vendor/shaperpoint.py, so it is solved for all users - if this is generic sharepoint problem (if yes, please provide link to some sharepoint specification if possible). 

btw do you know if there is any public instance this could be tested upon? or perhaps some dockerized target for integration tests? I have personally never used odata and Sharepoint together.

jenrik:
answer to question 1: adding the type overwrite to `vendor/sharepoint.py` would be more convenient, but but might be a gotcha. Input on how to activate the Sharepoint related overwrite are welcome.
I will see what I can find in terms of documentation, so for it's based on observations from the SharePoint application I have access to.

answer to question 2: Too my admittedly limited knowledge there is no public SharePoint instance we can use for testing, and SharePoint only run on Windows. There is couple of different version of SharePoint with different quirks, I believe the latest version is available in Microsoft Azure.

phanak-sap:
Hi @jenrik, will you have time in near future to finish on this PR? At the moment I cannot merge it to mainline, I cannot even sufficiently test it and existing tests are failing.

jfilak:
Nice PR! I would add some facade method overwritting and write extensive documentation with few examples.

jenrik:
sorry, I have abandoned this PR, I'm no longer in need of the functionality so I want have the time to work further on it

PR #243, opened 2023-03-08
Encode path option get and update

linda-sap:
Sometimes services expect the path to be percent encoded. This can especially be important 
when special variable types are key fields like Date type, where a ':' will appear in the path. 
In this case you can use an optional parameter to make the request encode the path.
Per default the variable encode_path is set to False.

### DIFF
--- a/docs/usage/deleting.rst
+++ b/docs/usage/deleting.rst
@@ -47,7 +47,7 @@ or by passing the EntityKey object
 Encode OData URL Path
 -------------------------------------------
 
-By default the paths of requests are percent encoded. However if this is not what you API expects, 
+By default the paths of requests are percent encoded. However if this is not what your API expects, 
 you can disable the encoding with the variable encode_path by setting it to False.
 
 
--- a/docs/usage/deleting.rst
+++ b/docs/usage/deleting.rst
@@ -42,3 +42,15 @@ or by passing the EntityKey object
     key = EntityKey(service.schema.entity_type('Employee'), ID=23)
     request = service.entity_sets.Employees.delete_entity(key=key)
     request.execute()
+
+
+Encode OData URL Path
+-------------------------------------------
+
+By default the paths of requests are percent encoded. However if this is not what you API expects, 
+you can disable the encoding with the variable encode_path by setting it to False.
+
+
+.. code-block:: python
+
+   request = service.entity_sets.Employees.delete_entity(key=key, encode_path=False)
--- a/docs/usage/deleting.rst
+++ b/docs/usage/deleting.rst
@@ -47,7 +47,7 @@ or by passing the EntityKey object
 Encode OData URL Path
 -------------------------------------------
 
-By default the paths of requests are percent encoded. However if this is not what your API expects, 
+By default the resource paths of requests are percent encoded. However if this is not what your API expects, 
 you can disable the encoding with the variable encode_path by setting it to False.
 
 
--- a/docs/usage/querying.rst
+++ b/docs/usage/querying.rst
@@ -157,14 +157,12 @@ these parameters by the method `custom(name: str, value: str)`.
 Encode OData URL Path
 -------------------------------------------
 
-Sometimes services expect the path to be percent encoded. This can especially be important 
-when special variable types are key fields like Date type, where a ':' will appear in the path. 
-In this case you can use an optional parameter to make the request encode the path.
-Per default the variable encode_path is set to False.
+By default the paths of requests are percent encoded. However if this is not what you API expects, 
+you can disable the encoding with the variable encode_path by setting it to False.
 
 .. code-block:: python
 
-    employee = northwind.entity_sets.Employees.get_entity(1, encode_path=True).execute()
+    employee = northwind.entity_sets.Employees.get_entity(1, encode_path=False).execute()
 
 (Experimental) Query server-side paginations using the __next field
 -------------------------------------------------------------------
--- a/docs/usage/querying.rst
+++ b/docs/usage/querying.rst
@@ -157,7 +157,7 @@ these parameters by the method `custom(name: str, value: str)`.
 Encode OData URL Path
 -------------------------------------------
 
-By default the paths of requests are percent encoded. However if this is not what your API expects, 
+By default the resource paths of requests are percent encoded. However if this is not what your API expects, 
 you can disable the encoding with the variable encode_path by setting it to False.
 
 .. code-block:: python
--- a/docs/usage/querying.rst
+++ b/docs/usage/querying.rst
@@ -154,7 +154,17 @@ these parameters by the method `custom(name: str, value: str)`.
 
     employees = northwind.entity_sets.Employees.get_entities().custom('sap-client', '100').custom('$skiptoken', 'ABCD').top(10).execute() 
     
+Encode OData URL Path
+-------------------------------------------
 
+Sometimes services expect the path to be percent encoded. This can especially be important 
+when special variable types are key fields like Date type, where a ':' will appear in the path. 
+In this case you can use an optional parameter to make the request encode the path.
+Per default the variable encode_path is set to False.
+
+.. code-block:: python
+
+    employee = northwind.entity_sets.Employees.get_entity(1, encode_path=True).execute()
 
 (Experimental) Query server-side paginations using the __next field
 -------------------------------------------------------------------
--- a/docs/usage/querying.rst
+++ b/docs/usage/querying.rst
@@ -157,7 +157,7 @@ these parameters by the method `custom(name: str, value: str)`.
 Encode OData URL Path
 -------------------------------------------
 
-By default the paths of requests are percent encoded. However if this is not what you API expects, 
+By default the paths of requests are percent encoded. However if this is not what your API expects, 
 you can disable the encoding with the variable encode_path by setting it to False.
 
 .. code-block:: python
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -442,10 +442,11 @@ class EntityGetRequest(ODataHttpRequest):
             urljoin(self._url, self.get_path(), '/$value'),
             connection,
             stream_handler)
-    
+
     def get_encode_path(self):
         """Getter for encode path flag"""
         return self._encode_path
+
 
 class NavEntityGetRequest(EntityGetRequest):
     """Used for GET operations of a single entity accessed via a Navigation property"""
@@ -615,7 +616,7 @@ class EntityModifyRequest(ODataHttpRequest):
 
     def get_default_headers(self):
         return {'Accept': 'application/json', 'Content-Type': 'application/json'}
-    
+
     def get_encode_path(self):
         """Getter for encode path flag"""
         return self._encode_path
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -371,7 +371,7 @@ class ODataHttpRequest:
 class EntityGetRequest(ODataHttpRequest):
     """Used for GET operations of a single entity"""
 
-    def __init__(self, handler, entity_key, entity_set_proxy, encode_path=False):
+    def __init__(self, handler, entity_key, entity_set_proxy, encode_path=True):
         super(EntityGetRequest, self).__init__(entity_set_proxy.service.url, entity_set_proxy.service.connection,
                                                handler)
         self._logger = logging.getLogger(LOGGER_NAME)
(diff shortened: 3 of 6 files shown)


jfilak:
@linda-sap  Welcome to the contributors. This is a nice pull request. Could you just explain why is the encoding optional? Isn't just a bug to not always encode it?

linda-sap:
@jfilak Thanks :) I currently can't think of a case where it shouldn't be encoded, rather just to not crash any existing use, where it might be expected unencoded. Would you opt for making it always encode or maybe to make the encoding the default?

jfilak:
@linda-sap yes, please swap the logic - make encoding default and offer the possibility to not encode for possible workarounds.

phanak-sap:
> @linda-sap Welcome to the contributors. This is a nice pull request. Could you just explain why is the encoding optional? Isn't just a bug to not always encode it?


> @linda-sap yes, please swap the logic - make encoding default and offer the possibility to not encode for possible workarounds.

Agree, it is a bugfix, but a for five years old part that may be a possible unforeseen breaking change for somebody. At least the workaround option for run of behavior that existed so far. Removal of this configuration option may be a good subtask for 2.0 version, right?

phanak-sap:
@linda-sap please add a changelog entry under the Unreleased as well, so it is not forgotten.

phanak-sap:
@linda-sap  The linter job did not run, but it is important for this PR as well as a gate-keeper. I have fixed it in PR #239. Please update your branch as well, so we do [not see this problem.](https://github.com/SAP/python-pyodata/actions/runs/4366078828/jobs/7640767611)

linda-sap:
@phanak-sap do I get you right, that you also prefer the option to make the encoding the default and leave the option to witch it off? I just want to be sure as also a good amount of tests need a change for this.

phanak-sap:
> @phanak-sap do I get you right, that you also prefer the option to make the encoding the default and leave the option to witch it off? I just want to be sure as also a good amount of tests need a change for this.

Yes. I would probably start more defensively exactly like you did :), but the @jfilak suggestion makes for sense for long future. Bugfix, not a new feature mentality. Encoding by default, with new parameter option to run old behavior for possible workarounds.

linda-sap:
@jfilak @phanak-sap I changed it so the encoding is done by default and added the encoding to delete and getEntitySet request. Also I adopted the tests. However I'm not sure what the best way is to include the encoding or not in the get_value() function. Whether the general ODataHttpRequest class should handle it.

phanak-sap:
Hi @linda-sap, please fix the linter build as well to be passing. You can run it locally as well, check "pylint" and "flake8": in the [contributing guide](https://github.com/SAP/python-pyodata/blob/master/CONTRIBUTING.md)

Two things from previous build were fixed, these remains. I am OK with `# pylint: disable` those rules for the EntityModifyRequest constructor. 

pyodata/v2/service.py:578: [R0902(too-many-instance-attributes), EntityModifyRequest] Too many instance attributes (8/7)
pyodata/v2/service.py:586: [R0913(too-many-arguments), EntityModifyRequest.__init__] Too many arguments (8/7)

docs/usage/querying.rst
```diff
@@ -154,7 +154,17 @@ these parameters by the method `custom(name: str, value: str)`.
 
     employees = northwind.entity_sets.Employees.get_entities().custom('sap-client', '100').custom('$skiptoken', 'ABCD').top(10).execute() 
     
+Encode OData URL Path
+-------------------------------------------
+
+Sometimes services expect the path to be percent encoded. This can especially be important
```
phanak-sap, line 160: delete the "especially" word. Just for readability, "especially" followed by "special variable" is unnecessary.

tests/test_service_v2.py
```diff
@@ -1563,6 +1606,77 @@ def test_batch_request(service):
     assert len(chset_response) == 1
     assert chset_response[0] is None   # response to update request is None
 
+@responses.activate
+def test_batch_request_with_decoded_path(service):
```
phanak-sap, line 1610: what does "decoded" means in this context? I understand the PR as that the batch can be "encoded path" or "not encoded path", I do not understand the "decoded" test case at all from its name.
Can you pls explain if I understand something wrong / rename if it is really a batch_request_not_encoded_path scenario?

tests/test_service_v2.py
```diff
@@ -1563,6 +1606,77 @@ def test_batch_request(service):
...
+        f'{URL_ROOT}/$batch',
+        body=response_body,
+        content_type='multipart/mixed; boundary=batch_r1',
+        status=202)
+
+    batch = service.create_batch('batch1')
+
+    chset = service.create_changeset('chset1')
+
+    employee_request = service.entity_sets.Employees.get_entity(23, encode_path=False)
```
phanak-sap, line 1651: I was able to find only `encode_path=False` scenario for a batch payload. I would expect at least some encode_path=False scenarios for direct, non-batch URLs as well, all other tests used the default True and wrap the resource path vie `quote()`. Does not need to be copy paste of every modified test , but perhaps adding of one new test - ideallz parametrized, with various problematic resource paths based on different types (string, date etc.)

phanak-sap (commented):
I have small nitpick requests, but generally speaking, I am OK with this feature to be merged and released as is.

tests/test_service_v2.py
linda-sap: Yes, ..not_encoded is a better naming. I changed it, Thank you!

tests/test_service_v2.py
linda-sap: There is one test which tests delete and one for get requests with not encoding the path.
[test_delete_entity_not_encoded_path](https://github.com/linda-sap/python-pyodata/blob/1e6e1a982ea6ee8b16d32058526d7fc7fd7d11c5/tests/test_service_v2.py#L661)
[test_get_entity_not_encoded_path](https://github.com/linda-sap/python-pyodata/blob/1e6e1a982ea6ee8b16d32058526d7fc7fd7d11c5/tests/test_service_v2.py#L1478)

Is this sufficient for you?

docs/usage/deleting.rst
```diff
@@ -41,4 +41,16 @@ or by passing the EntityKey object
...
     request = service.entity_sets.Employees.delete_entity(key=key)
-    request.execute()
\ No newline at end of file
+    request.execute()
+
+
+Encode OData URL Path
+-------------------------------------------
+
+By default the paths of requests are percent encoded. However if this is not what you API expects,
```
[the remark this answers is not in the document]
linda-sap: Adjusted, Thanks

linda-sap:
@phanak-sap thanks a lot for your feedback!
Just so we are on the same page, the gut_value() functions (e.g. in [test_entity_get_value_with_proxy](https://github.com/linda-sap/python-pyodata/blob/1e6e1a982ea6ee8b16d32058526d7fc7fd7d11c5/tests/test_service_v2.py#L1217) and [test_entity_get_value_1on1_with_proxy](test_entity_get_value_with_proxy)) do not return an object with an encoded path as those are not the requests to be sent.

phanak-sap:
> @phanak-sap thanks a lot for your feedback! Just so we are on the same page, the gut_value() functions (e.g. in [test_entity_get_value_with_proxy](https://github.com/linda-sap/python-pyodata/blob/1e6e1a982ea6ee8b16d32058526d7fc7fd7d11c5/tests/test_service_v2.py#L1217) and [test_entity_get_value_1on1_with_proxy](test_entity_get_value_with_proxy)) do not return an object with an encoded path as those are not the requests to be sent.

I do not understand "those are not the requests to be sent." correctly perhaps. Are we talking about method that basically add  `$value` as a query parameter? That is different part of URL string.

Maybe this part of documentation will tie better the odata documentation and this library:
https://github.com/SAP/python-pyodata/blob/5a232b5013ce16c81996409c19d17311c7434203/docs/usage/urls.rst
https://www.odata.org/documentation/odata-version-2-0/uri-conventions/

pyodata/v2/service.py
```diff
@@ -545,36 +552,45 @@ def set(self, **kwargs):
...
     """Used for modyfing entities (UPDATE/MERGE operations on a single entity)
 
        Call execute() to send the update-request to the OData service
        and get the modified entity."""
 
     ALLOWED_HTTP_METHODS = ['PATCH', 'PUT', 'MERGE']
-
-    def __init__(self, url, connection, handler, entity_set, entity_key, method="PATCH"):
+    # pylint: disable=too-many-arguments
+    def __init__(self, url, connection, handler, entity_set, entity_key, method="PATCH", encode_path=True):
```
phanak-sap, line 587: pyodata/v2/service.py:587:5: E301 expected 1 blank line, found 0
make: *** [Makefile:35: lint] Error 1

from linter build, 
run locally with `flake8 --config=.flake8  pyodata`

phanak-sap: To satisfy pylint and flake 8, it is enough to add new line, like this:

```
    ALLOWED_HTTP_METHODS = ['PATCH', 'PUT', 'MERGE']

    # pylint: disable=too-many-arguments
    def __init__(self, url, connection, handler, entity_set, entity_key, method="PATCH", encode_path=True)
```:

The PR was not allowed with commits from maintainers to the branch, @linda-sap pls update on your end. 

Otherwise the PR is looking good. If with no communication for a long time, I will merge it as is and fix in subsequent commit.

linda-sap:
> > @phanak-sap thanks a lot for your feedback! Just so we are on the same page, the gut_value() functions (e.g. in [test_entity_get_value_with_proxy](https://github.com/linda-sap/python-pyodata/blob/1e6e1a982ea6ee8b16d32058526d7fc7fd7d11c5/tests/test_service_v2.py#L1217) and [test_entity_get_value_1on1_with_proxy](test_entity_get_value_with_proxy)) do not return an object with an encoded path as those are not the requests to be sent.
> 
> I do not understand "those are not the requests to be sent." correctly perhaps. Are we talking about method that basically add `$value` as a query parameter? That is different part of URL string.
> 
> Maybe this part of documentation will tie better the odata documentation and this library: https://github.com/SAP/python-pyodata/blob/5a232b5013ce16c81996409c19d17311c7434203/docs/usage/urls.rst https://www.odata.org/documentation/odata-version-2-0/uri-conventions/

Okay, I just wanted to make sure there is no need to also encode the url in these methods.

phanak-sap:
> > > @phanak-sap thanks a lot for your feedback! Just so we are on the same page, the gut_value() functions (e.g. in [test_entity_get_value_with_proxy](https://github.com/linda-sap/python-pyodata/blob/1e6e1a982ea6ee8b16d32058526d7fc7fd7d11c5/tests/test_service_v2.py#L1217) and [test_entity_get_value_1on1_with_proxy](test_entity_get_value_with_proxy)) do not return an object with an encoded path as those are not the requests to be sent.
> > 
> > 
> > I do not understand "those are not the requests to be sent." correctly perhaps. Are we talking about method that basically add `$value` as a query parameter? That is different part of URL string.
> > Maybe this part of documentation will tie better the odata documentation and this library: https://github.com/SAP/python-pyodata/blob/5a232b5013ce16c81996409c19d17311c7434203/docs/usage/urls.rst https://www.odata.org/documentation/odata-version-2-0/uri-conventions/
> 
> Okay, I just wanted to make sure there is no need to also encode the url in these methods.

Well, `.get_path()` is now encoded in your PR, right? Rest three methods cover different URL parts, out of scope of this PR.  And the `get_value()` from original question was resolved that it adding query parameter, so refers to different url part. 

So I am under impression that everything is covered and this is just "communication noise" :). I will wait some time for possible reaction, but I will merge this today or tomorrow.

linda-sap:
> > > > @phanak-sap thanks a lot for your feedback! Just so we are on the same page, the gut_value() functions (e.g. in [test_entity_get_value_with_proxy](https://github.com/linda-sap/python-pyodata/blob/1e6e1a982ea6ee8b16d32058526d7fc7fd7d11c5/tests/test_service_v2.py#L1217) and [test_entity_get_value_1on1_with_proxy](test_entity_get_value_with_proxy)) do not return an object with an encoded path as those are not the requests to be sent.
> > > 
> > > 
> > > I do not understand "those are not the requests to be sent." correctly perhaps. Are we talking about method that basically add `$value` as a query parameter? That is different part of URL string.
> > > Maybe this part of documentation will tie better the odata documentation and this library: https://github.com/SAP/python-pyodata/blob/5a232b5013ce16c81996409c19d17311c7434203/docs/usage/urls.rst https://www.odata.org/documentation/odata-version-2-0/uri-conventions/
> > 
> > 
> > Okay, I just wanted to make sure there is no need to also encode the url in these methods.
> 
> Well, `.get_path()` is now encoded in your PR, right? Rest three methods cover different URL parts, out of scope of this PR. And the `get_value()` from original question was resolved that it adding query parameter, so refers to different url part.
> 
> So I am under impression that everything is covered and this is just "communication noise" :). I will wait some time for possible reaction, but I will merge this today or tomorrow.

Correct, the get_path() is encoded and the others not.

phanak-sap:
Thank you for all the work.

Issue #250, opened 2023-05-15
ci: fix test failures - ChunkedEncodingError


phanak-sap:
https://github.com/SAP/python-pyodata/actions/runs/4740338469/jobs/8898438292

Complete job name: build (3.10, 4.9.1)

=========================== short test summary info ============================
FAILED tests/test_service_v2.py::test_function_import_without_return_type_wrong_code - requests.exceptions.ChunkedEncodingError: ('Connection broken: IncompleteRead(10 bytes read, -10 more expected)', IncompleteRead(10 bytes read, -10 more expected))
FAILED tests/test_service_v2.py::test_update_entity - requests.exceptions.ChunkedEncodingError: ('Connection broken: IncompleteRead(86 bytes read, -86 more expected)', IncompleteRead(86 bytes read, -86 more expected))
======================== 2 failed, 265 passed in 5.40s =========================
make: *** [Makefile:39: test] Error 1
Error: Process completed with exit code 2.

phanak-sap:
This is a older build. After sync with master of my own fork (just readme update), now all compatibility matrix builds are failing

https://github.com/phanak-sap/python-pyodata/actions/runs/4972829981

Builds which were passing used version urllib3-1.26.15, while failing builds are now using urllib3-2.0.2 - from dev-dependencies.txt

PR #251, opened 2023-05-15
tests: fix failures from bump of transitive deps to urllib3 2.X

phanak-sap:
- remove remove test_function_import_without_return_type_wrong_code

The entire logic of the test was based on scenario that server returns response of HTTP 204 No Content (which ends with HTTP Header) with some content in HTTP Body, which is breaking the HTTP protocol logic itself.

The Responses package generated such weird simulated response and was parsed with urllib 1.X in previous versions of Requests, but now, with being compatible with urllib3 2.x, breaking changes were introduced within this transitive dependency.

On the requests side, the bump to same urllib3 2.x means test failures with error:

requests.exceptions.ChunkedEncodingError: ('Connection broken: IncompleteRead(10 bytes read, -10 more expected)', IncompleteRead(10 bytes read, -10 more expected))

- test_update_entity

Same rationale, removed invalid body for expected HTTP 204 simulated response

Additional info:

getsentry/responses#636
https://urllib3.readthedocs.io/en/latest/v2-migration-guide.html

RFC 2616  - HTTP 1.1 - 10.2.5 - 204 No Content
https://datatracker.ietf.org/doc/html/rfc2616#section-10.2.5

### DIFF
--- a/tests/test_service_v2.py
+++ b/tests/test_service_v2.py
@@ -464,27 +464,6 @@ def test_function_import_without_return_type_wrong_code(mock_warning, service):
 
 
 @responses.activate
-@patch('logging.Logger.warning')
-def test_function_import_without_return_type_wrong_code(mock_warning, service):
-    """A simple function call without return type should not return any data"""
-
-    # pylint: disable=redefined-outer-name
-
-    responses.add(
-        responses.GET,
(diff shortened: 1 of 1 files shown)


phanak-sap:
closes #250

PR #259, opened 2023-11-15
Fix adding expanded entities to EntityProxy

### LINKED ISSUE #258: Expanded properties don't have their own results wrapper in V2 like this library expects
In OData V2 JSON, expanded properties are simple arrays of the contained entities. 

See https://www.odata.org/documentation/odata-version-2-0/json-format/
Section 9, 9. Inline Representation of Associated Entries

However, pyodata expects the nested entities to be wrapped in a results object: 

https://github.com/SAP/python-pyodata/blob/ea558f504532e378e182952db5e78aead3d23a6e/tests/test_service_v2.py#L1492C18-L1492C25
https://github.com/SAP/python-pyodata/blob/ea558f504532e378e182952db5e78aead3d23a6e/pyodata/v2/service.py#L875C31-L875C31

This is not according to spec and will fail with common implementations (including SAP service providers) as in  the expanded property will discard the array received from the service and will default to an empty list.

manuelseeger:
EntityProxy expected expanded properties to be wrapped in a 'results' object, which isn't to V2 JSON spec. This fix removes this expectation. The original behavior is not removed, so if there happens to be an expanded property wrapped in a 'results' object, EntityProxy continues to work as before.

Fixes #258

### DIFF
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -873,9 +873,11 @@ class EntityProxy:
                         # if there are no entities available, received data consists of
                         # metadata properties only.
                         if 'results' in proprties[prop.name]:
-
                             # available entities are serialized in results array
                             for entity in proprties[prop.name]['results']:
+                                self._cache[prop.name].append(EntityProxy(service, None, prop_etype, entity))
+                        else:
+                            for entity in proprties[prop.name]:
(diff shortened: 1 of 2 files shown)


phanak-sap:
Simple, yet backward compatible fix with necessary changes of test suite. Perfect. Thanks  @manuelseeger

PR #270, opened 2024-01-21
ci: extend compatibility checks for lxml 5.x

phanak-sap:


phanak-sap:
It is a bit surprising, but LXML 5.x is better compatible with older python versions than previous 4.x releases. Nice.

PR #273, opened 2024-05-20
chore: bump linters in dev-requirements.txt

phanak-sap:
- bump pylint, flake and bandit to current version
- set version for sphinx


phanak-sap:
dropping, this obsolete and bump to new version of pylint is already seen locally as significant chuck of work

Issue #280, opened 2024-07-08
Chaining FilterExpressions


developer992:
Hello,

Trying to follow this PR: https://github.com/SAP/python-pyodata/pull/113

love this change, very good implementation but am having some problems chaining AND and OR operators together.

```
item = {
    "id": 1,
    "name": 'item name',
    "description": "foo-bar",
    "city": "1000",
}

# Query should look like this:
term="foo"      # search by string term on two fields - OR on name and description
city=1000       # AND search only those with city = 1000


get_entities().filter(Q(name__contains="foo") | Q(description__contains="foo"), city="1000")

# It does this:
{'$filter': ["city eq '1000' and (substringof('foo', name) eq true) or (substringof('foo', description) eq true)"]}
```

But it needs another ( ) around substring filters to fetch correct items.

Is this possible using these filters? 

Many thanks!

developer992:
**wait, I GOT IT** 

this works:
```
get_entities().filter(Q(name__contains="foo") | Q(description__contains="foo"), Q(city="1000"))
```

Not sure if it's a bug or just documentation wrong.

Anyway, thanks!

developer992:
Actually, no

I think this is a bug:

The problem is in this function:

```
def _combine_expressions(self, expressions):
    return ' and '.join(expressions)
```

This function gets called several times during the process and the problem is only in the last call when it calls it with all expressions.

I've improved it like so:

Basically it wraps expressions in "( )"

```
    def _combine_expressions(self, expressions):
        if len(expressions) > 1:
            combined = ""
            for counter, expr in enumerate(expressions):
                combined += f"{' and ' if counter > 0 else ''}({expr})"

            return combined
        
        return expressions[0]
```

in then calls the service with such url:

/v2/service/odata-service/odata-table?%24top=100&%24skip=0&%24filter=%28plant+eq+%271000%27%29+and+%28h5+eq+%279%27+or+h5+eq+%27VSA113211%27+or+h5+eq+%27EVL110022%27%29+and+%28mr+eq+%27N40%27%29&%24inlinecount=allpages

which seems to return the correct items.

I hope i haven't broke anything else though.

phanak-sap:
Hi @developer992  - seems you put lot of time and effort into this. 

[...]

PR #113, opened 2020-06-28
Django style filtering for GetEntitySetRequest

bartonip:
The existing filtering mechanism is not very intuitive, requiring interpolation of OData queries or using the convoluted boolean logic syntax of `GetEntitySetFilter` to achieve complex queries. 

I have implemented a Django-like filtering system that allows filtering of EntitySets in a much more intuitive way.

**Syntax**
---
For example, a simple equivalency query was previously written like this:
```
request = service.entity_sets.Employees.get_entities().filter("FirstName eq 'Tim' and LastName eq 'Smith'")
```

or even more confusingly: 
```
request = service.entity_sets.Employees.get_entities().filter(sef.and_(
                                                             request.FirstName == 'Tim',
                                                             request.LastName == 'Smith'))
```

With this proposed update, future calls will look like this.

```
request = service.entity_sets.Employees.get_entities().filter(FirstName="Tim", LastName="Smith")
```

Tell me that's not way more beautiful. ;)

**Filter Expressions**
---

Next, I have implemented some lookups using Django-like syntax that opens up some more advanced query scenarios.
The operators `lt`, `lte`, `gt`, `gte`, `in`, `contains`, `startswith`, `endswith`, `length` and `range` have been implemented.

Examples below:
```
request = service.entity_sets.Employees.get_entities().filter(ID__lte=20)

request = service.entity_sets.Employees.get_entities().filter(ID__gt=20)

request = service.entity_sets.Employees.get_entities().filter(ID__in=[20, 21, 22])

request = service.entity_sets.Employees.get_entities().filter(FirstName__contains="Tim")

request = service.entity_sets.Employees.get_entities().filter(FirstName__length=3)

```

There is also the introduction of the `FilterExpression` class which is equivalent to Django's Q class.

This allows common queries to be stored as objects:
```
from pyodata.v2.service import FilterExpression as Q
company = Q(CompanyID=1234)

request = service.entity_sets.Employees.get_entities().filter(company, FirstName="Tim", LastName__startswith="S")

```

This also allows the `or` operator to be used to or statements:
```
request = service.entity_sets.Employees.get_entities().filter(Q(FirstName="Tim") | Q(FirstName="John"), LastName="Smith))
```

All changes are backwards compatible and do not disrupt existing implementations of this library, and upgrading to this syntax should be simple if not trivial.

PR #283, opened 2024-09-11
Bugfix when chaining expressions using OR

developer992:
- Wraps expressions in brackets when using more than one expression
so this
`/Employees/$count?$filter=%28ID%20eq%2023%20and%20NickName%20eq%20%27Steve%27%29%20or%20%28ID%20eq%2025%20and%20NickName%20eq%20%27Tim%27%29` 

`(ID eq 23 and NickName eq 'Steve') or (ID eq 25 and NickName eq 'Tim')`

becomes 
`/Employees/$count?$filter=%28%28ID%20eq%2023%29%20and%20%28NickName%20eq%20%27Steve%27%29%29%20or%20%28%28ID%20eq%2025%29%20and%20%28NickName%20eq%20%27Tim%27%29%29`

`((ID eq 23) and (NickName eq 'Steve')) or ((ID eq 25) and (NickName eq 'Tim'))`

- adds a test
- fixes existing tests
- addresses issue https://github.com/SAP/python-pyodata/issues/280

Example 2:
consider this example, it demonstrates a bit better

`startswith(NameFirst, 'Steve') eq true and ID eq 23 or ID eq 25 or ID eq 28` becomes  `(startswith(NameFirst, 'Steve') eq true) and (ID eq 23 or ID eq 25 or ID eq 28)`

### DIFF
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -1268,7 +1268,14 @@ class GetEntitySetFilterChainable:
 
     # pylint: disable=no-self-use
     def _combine_expressions(self, expressions):
-        return ' and '.join(expressions)
+        if len(expressions) > 1:
+            combined = ""
+            for counter, expr in enumerate(expressions):
+                combined += f"{' and ' if counter > 0 else ''}({expr})"
+
+            return combined
+
+        return expressions[0]
 
     # pylint: disable=too-many-return-statements, too-many-branches
     def _build_expression(self, field_name, operator, value):
--- a/tests/test_service_v2.py
+++ b/tests/test_service_v2.py
@@ -2541,12 +2541,32 @@ def test_count_with_chainable_filter_multiple(service):
 
     responses.add(
         responses.GET,
-        f"{service.url}/Employees/$count?%24filter=ID%20eq%2023%20and%20NickName%20eq%20%27Steve%27",
+        f"{service.url}/Employees/$count?%24filter=%28ID%20eq%2023%29%20and%20%28NickName%20eq%20%27Steve%27%29",
         json=3,
         status=200)
 
     employees = service.entity_sets.Employees.get_entities()
     request = employees.filter(ID=23, NickName="Steve").count()
+
+    assert isinstance(request, pyodata.v2.service.GetEntitySetRequest)
+
+    assert request.execute() == 3
+
+
+@responses.activate
+def test_chaining_fix(service):
+    from pyodata.v2.service import FilterExpression as Q
+
+    url = f"{service.url}/Employees?%24filter=%28startswith%28NameFirst%2C+%27Steve%27%29+eq+true%29+and+%28ID+eq+23+or+ID+eq+25+or+ID+eq+28%29"
+    
+    responses.add(
+        responses.GET,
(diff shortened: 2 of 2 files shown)


developer992:
Hi,

couple of notes

I couldn't get pylint to work, i get error:

`AttributeError: module 'collections' has no attribute 'MutableMapping'`

Must be some version mismatch but I couldn't (yet) track it down.

I added a test and fixed a few existing ones but i think these tests aren't the best. These basically just verify the constructed URL is correct and not actually test the remote service.

There could be a better way for this using [pytest-vcr](https://pypi.org/project/pytest-vcr/) which does exactly that. Test remote services and their responses. But this would require a bit more effort to implement, if you all want to go this route.

Anyway, i probably forgot other things, let me know what you guys think.

phanak-sap:
Hi @developer992

Thanks for all your work. Greatly appreciated.

I will look at this more deeply soon. 

1. The linter was probably local problem, even that the version is fixed in the dev-requirements.txt, but it could be affected by other things (python version if nothing else). It passed in the PR actions build, so no worries. 

2. For this PR, new test code works, but I would like to return to the point 4 from my comment in the original issue - https://github.com/SAP/python-pyodata/issues/280#issuecomment-2276607288. It is basically the same problem that the test is passing, because the fix matches the hardcoded URL in the new test.. but is this one URL the only valid for the chained expression in the python code?


from the issue comment:
```
filter1 eq 'foo' and ( filter2 eq true or filter3 eq 'bar' )
or
(filter1 eq 'foo' and  filter2 eq true) or filter3 eq 'bar' 
```
both should most probably return different set of data in the response.

You responded in the middle with the _but we'll know more after running the tests :)_ But there is just one quite simple test added.

I am not sure if you are covering all possible groupings or pre-selected one grouping only for the ORM style - and any other complex filter query would have to be written directly in the standard pyodata way.

This is potentially breaking change, so if you are using the fixed version from your fork for several weeks, may I ask to add few more tests for these more complex chained expression that would use the experience you gained? This is moving in the right direction overall and I would love to merge it and release it - I have planned it for the next bugfix release. So this is not comment discarding the PR, on the contrary - request for improving a bit more.

3. You may also consider improving the documentation (this PR or perhaps second one, after this is merged) with your experience, not just tests. As you already know, the PR #113 is a community feature and not what was considered "good enough minimum " pyodata in the start, so if you see benefit in writing queries this ORM way, you can "clean the path" for anybody following you.
https://github.com/SAP/python-pyodata/blob/master/docs/usage/querying.rst#get-entities-matching-a-complex-filter-in-orm-style

4. Last but not least.. update CHANGELOG.md with your fix and your name under https://github.com/SAP/python-pyodata/blob/master/CHANGELOG.md. It is your fix after all.

developer992:
i am not sure howto handle this

it works for us, so i'll keep it but if it depends on remote service, then i'm lost on how to continue


thanks anyway tho, you may delete the PR

phanak-sap:
I am closing this as a part of cleanup of old, stale PRs. The original problem remains valid, this PR tackled it in a backward-compatibility breaking change.  See comment there.

PR #284, opened 2024-09-19
Escape slashes in entity keys when they appear in the URL path

### LINKED ISSUE #282: Incorrect escaping of path
Greetings,

pyodata does not correctly escape the path when an entity key contains a `/`.  For example SAP's OData service for Software Components retrieves an SC called `ZFOO` with the URL:
```
https://hostname/sap/opu/odata/sap/A4C_A2G_GHA_SC/SoftwareComponents%28%27ZFOO%27%29
```
But if I try to get an SC called `/BAZ/FOO` the URL is malformed as:
```
https://hostname/sap/opu/odata/sap/A4C_A2G_GHA_SC/SoftwareComponents%28%27/BAZ/FOO%27%29
```
but the `/` characters ought to be escaped as 
```
https://hostname/sap/opu/odata/sap/A4C_A2G_GHA_SC/SoftwareComponents%28%27%2FBAZ%2FFOO%27%29
```

Looking at the pyodata code I can see that `EntityGetRequest.get_path()` is calling `quote()` which by default considers `safe='/'`.  It seems to me it should override `safe=''` so that `/` characters are escaped.  I am new to the pyodata code so I'm not sure if this issue might also lurk on operations other than `EntityGetRequest`.  I'll appreciate any feedback or expansion on my analysis.

frij-aws:
Pyodata does not correctly escape the path when an entity key contains a /. This minor correction quotes the entity key with `safe=""` to ensure slashes are quoted.  The rest of the path is quoted with the default (`safe='/') so that path delimiters are not quoted.

Closes #282

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Fixed
-- Escape slashes in entity keys when they appear in the URL path
+- Escape slashes in entity keys when they appear in the URL path - Jon Friesen
 
 ### Removed
 - Python 3.7 (long after its EOL) is no longer supported by pyodata. Python 3.8 is now minimal supported version. - Petr Hanak
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
-
+### Fixed
+- Escape slashes in entity keys when they appear in the URL path
 
 ### Removed
 - Python 3.7 (long after its EOL) is no longer supported by pyodata. Python 3.8 is now minimal supported version. - Petr Hanak
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -405,7 +405,7 @@ class EntityGetRequest(ODataHttpRequest):
 
     def get_path(self):
         if self.get_encode_path():
-            return quote(self._entity_set_proxy.last_segment + self._entity_key.to_key_string())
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,7 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
-
+### Fixed
+- Escape slashes in entity keys when they appear in the URL path
```
phanak-sap, line 9: Note: if you see the changelog entries, we add also author names for important changes. So feel free to add yourself to this line :)

phanak-sap:
HI @frij-aws
 
thank you for your work. Generic stuff I commented in the issue, here just what's relevant to the code changes.  Nothing existing is changed or broken, that's great. 

If I may, I would like to ask you to enhance bit the test code.

- up to consideration for you is whether it would make sense to you to update existing metadata.xml or create new metadata.xml for the test fixtures with these "interesting" entity names. But the new tests are good enough for covering this.
- however I would like you to also add new test for this entity id case related to batch processing. I do not have service at hand with such $metadata I could test this PR against directly and I am not sure if this fix needs to be applied in the encode_multipart / decode_multipart functions as well. So pls check in the test_service_v2.py `test_batch_request` and `test_batch_request_not_encoded_path ` and add new one that will cover the expected encoding for your `FOO/BAR` entity.

phanak-sap:
Since you added additional PR #292 -  do you have somewhere the intended significant refactor" you see ?  Any results for the "some time to test some more scenarios to confirm if the workaround works in all cases" ?

What do you thing about the work from this pull request now?

frij-aws:
Let's cancel this PR.  It will be too hard to implement fully and the use of `[encode_path=False](https://github.com/SAP/python-pyodata/pull/encode_path) solves the problem for me.

PR #292, opened 2025-04-01
Adding two missing possibilities for ValueListParameter

frij-aws:
Adding two missing possibilities for `ValueListParameter` from from the [SAP common vocabulary list](https://sap.github.io/odata-vocabularies/vocabularies/Common.html)


phanak-sap:
Hi [frij-aws](https://github.com/frij-aws).

Thx for pull request. Straightforward and simple. 

1) I am quite surprised no test was actually need to be updated as well - clearly missing some automation here :)

2) Am I correct that we have in pyodata already hardcoded things that are SAP specific in a model.py, that tries to be covering Odata specification and adding SAP-only things ideally to different, isolated module like vendors/SAP.py? :)

Both are just open question, I am willing to merge as is.

PR #298, opened 2026-04-17
Return a list of EntityProxies when FunctionImport return type is a Collection

Ixaruz:
What the title says.

I had to  rename the `EntitySet` of the `FunctionImport` in the test metadata.xml to not include the namespace, or else the constructor of `EntityProxy` would be unable to find the right entity set...

since the `EntitySet` in the `FunctionImport` of `get_max` also was just `"TemperatureMeasurements"` (not `"EXAMPLE_SRV.TemperatureMeasurements"`), I assume it was just an oversight in the xml (as it wasn't tested before).

I don't know if my "fix" approach is too short sighted, I just wanted to provide potential fixes, as we needed them in our production.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- service: let FunctionRequests return a list of EntityProxies instead of the raw json, when the `ReturnType` is a Collection. - Emil B.
+
 ## [1.11.2]
 
 ### Fixed
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -1712,6 +1712,15 @@ class FunctionContainer:
                 entity_set = self._service.schema.entity_set(fimport.entity_set_name)
                 return EntityProxy(self._service, entity_set, fimport.return_type, response_data)
 
+            # 1.b alternatively, if return type is a Collection, return a list of appropriate entity proxy
+            if isinstance(fimport.return_type, model.Collection):
+                entity_set = self._service.schema.entity_set(fimport.entity_set_name)
+                collection_item_type = fimport.return_type.item_type
+                collection = []
+                for entity in response_data['results']:
+                    collection.append(EntityProxy(self._service, entity_set, collection_item_type, entity))
+                return collection
+
             # 2. return raw data for all other return types (primitives, complex types encoded in dicts, etc.)
             return response_data
 
--- a/pyodata/v2/service.py
+++ b/pyodata/v2/service.py
@@ -1714,9 +1714,16 @@ class FunctionContainer:
 
             # 1.b alternatively, if return type is a Collection, return a list of appropriate entity proxy
             if isinstance(fimport.return_type, model.Collection):
+                total_count = None
+                next_url = None
+                if '__count' in response_data:
+                    total_count = int(response_data['__count'])
+                if '__next' in response_data:
+                    next_url = response_data['__next']
+                collection = ListWithTotalCount(total_count, next_url)
+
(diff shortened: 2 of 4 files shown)


phanak-sap (changes requested):
Hi @Ixaruz 

You definitely hit a unfinished implementation and thanks for submitting a PR directly instead of just issue. 

I was away on vacation and then having post-vacation more intense work, so had time just now to look at this.

The PR is going to right place with handling Collections as return type. However, comparing with odata v2 spec and current implementation from PR #188 there is some information lost in current PR.

**Before this PR:** FunctionImport with Collection return type returned raw JSON, 
so users could still access `response_data['__next']` manually.

**After this PR:** Only `results` array is extracted; `__next` is discarded.

**Suggested fix:** Return `ListWithTotalCount` (like `get_entities()` does) instead of plain `list`:  

Same problem is for `__count`, which was possible to be present in raw response so far .`__count` can be present in Collection responses when $inlinecount=allpages was requested

[OData v2 spec JSON format](https://www.odata.org/documentation/odata-version-2-0/json-format/)
[Odata v2 spec inlinecount](https://www.odata.org/documentation/odata-version-2-0/uri-conventions/#InlinecountSystemQueryOption)

I have generated with Claude sample test that you may include into PR that should expose this problem in PR code. 

```
@responses.activate
def test_function_import_collection_with_pagination_metadata(service):
    """Function call with collection return type including pagination metadata
    
    This test demonstrates that both __next and __count metadata should be preserved
    when a FunctionImport returns a partial Collection with inline count.
    Per OData v2 spec, this metadata is part of the Collection response format.
    """
    # pylint: disable=redefined-outer-name

    responses.add(
        responses.GET,
        f'{service.url}/get_best_measurements',
        headers={'Content-type': 'application/json'},
        json={'d': {
            'results': [{
                'Sensor': 'sensor1',
                'Date': "/Date(1776417069000)/",
                'Value': '29.8d'
            }, {
                'Sensor': 'sensor2',
                'Date': "/Date(1776417578000)/",
                'Value': '30.2d'
            }],
            '__count': '50',
            '__next': 'https://example.com/service/get_best_measurements?$skiptoken=2'
        }},
        status=200)

    result = service.functions.get_best_measurements.execute()
    
    assert isinstance(result, list)
    assert len(result) == 2
    assert isinstance(result[0], pyodata.v2.service.EntityProxy)
    assert isinstance(result[1], pyodata.v2.service.EntityProxy)
    
    # FAIL: Collection should be ListWithTotalCount to preserve pagination metadata
    # This matches the behavior of EntitySetRequest.get_entities()
    assert hasattr(result, 'total_count'), \
        "FunctionImport Collection should preserve __count like get_entities()"
    assert result.total_count == 50
    
    assert hasattr(result, 'next_url'), \
        "FunctionImport Collection should preserve __next like get_entities()"
    assert result.next_url == 'https://example.com/service/get_best_measurements?$skiptoken=2'
```

Ixaruz:
Thank you for giving me such a detailed review and guidance.
I hope you find the changes satisfactory.

And sorry for the wait as well.

phanak-sap (changes requested):
Hi, looks good, just pls add disable for pylint rule too-many-branches that makes currently the build failing.

I will need to change things around the _next method so no problem to leave as is, yes, more branching, but still readable and probably will be refactored anyway

```
pyodata/v2/service.py:1662: [R0912(too-many-branches), FunctionContainer.__getattr__.function_import_handler] Too many branches (15/12)

Error: Process completed with exit code 2.
```

phanak-sap (approved):
lgtm as is, some small notes will be addressed directly by me in follow-up PR so this does not drag too much long.

PR #305, opened 2026-06-20
service, vendor.SAP: add response_hook for pre-handler response inspection

phanak-sap:
- Add an optional response_hook=None parameter to Service and Client

  response_hook fires inside _call_handler() before the domain handler runs, covering both execute() and async_execute(). The hook is a stateless Callable[[response], None]; raising from it propagates to the caller and suppresses the domain result. No HTTP networking objects cross the OData API boundary.

- Add sap_header_error_hook() to pyodata.vendor.SAP

  reads the sap-message response header and raises BusinessGatewayError when severity is "error".

- Update tests and documentation accordingly

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -336,7 +336,7 @@ exclude-protected=_asdict,_fields,_replace,_source,_make
 [DESIGN]
 
 # Maximum number of arguments for function / method
-max-args=7
+max-args=8
 
 # Argument names that match this expression will be ignored. Default to name
 # with leading underscore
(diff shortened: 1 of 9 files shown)


phanak-sap:
Left as open question for reviewers:

 - does that cover the original use case that was workerounded elsewhere? Is it usable together with sapcli or anything else
 - is 1 hook only OK or bad and should it be instead be list of hooks with possibility to register multiple different stateless response hooks? The logic chaining can always be solved IMHO outside of pyodata and provide one hook only for enforced simplicity.

filak-sap:
Replaces the original PR #154

filak-sap (approved):

PR #307, opened 2026-07-08
tests: fix failing test in MockHeaderResponse

phanak-sap:
bytes is the correct concrete type for the content field and matches the b'' default value.


phanak-sap:
last two PRs were passing independently, but after putting them together in master, new test failure raised in CI.

filak-sap (approved):

PR #308, opened 2026-07-09
service: guard against cross-origin __next URL redirection

phanak-sap:
OData servers can legitimately paginate across path segments, but the
  __next value is server-controlled and untrusted. Validate that its origin
  (scheme + host + port) matches the configured service root before
  dispatching, so session-level credentials are never forwarded to a host
  the application did not configure. Raises PyODataException on mismatch.

  Fix is applied in ODataHttpRequest._build_request(), the single point
  both execute() (sync) and async_execute() (async) pass through, so one
  check covers all dispatch paths. Depends on stdlib urllib.parse only,
  keeping pyodata networking-library agnostic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
-- service: `response_hook` parameter, enables inspection or rejection of raw responses (e.g. header-encoded SAP domain errors) without leaking HTTP transport objects through the OData API boundary.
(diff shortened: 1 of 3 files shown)


phanak-sap:
Build will currently fail because PR #307 not yet merged.

Important should be passing of both newly added tests (passing localy)

`pytest tests/test_service_v2.py::test_next_url_cross_origin_raises tests/test_service_v2.py::test_next_url_same_origin_allowe`d

filak-sap (changes requested):
Just do not open PRs if you know that tests are failing and another fixing the tests is pending please! Now merge the tests fixing PR and rebase this one.

phanak-sap:
> Just do not open PRs if you know that tests are failing and another fixing the tests is pending please! Now merge the tests fixing PR and rebase this one.

sure, that was indeed the intended path forward for me.

filak-sap (approved):

PR #312, opened 2026-07-28
service: resolve relative __next URLs against the service root

phanak-sap:
OData v2 permits relative __next values. Previously these were passed verbatim into the cross-origin check, which compared an empty scheme/netloc against the service origin and always raised PyODataException.

Now _build_request() detects a relative __next (no scheme, no netloc) and resolves it via urljoin(self._url, ...) before the origin check. Protocol- relative URLs (//host/...) are treated as absolute and validated as before.

Tests added:
- relative __next resolves to the correct absolute URL and dispatches
- protocol-relative __next pointing to a foreign host is still rejected


phanak-sap:
This is a follow-up feature regression fix for #308.

pyodata/v2/service.py
```diff
@@ -297,13 +297,19 @@ def add_headers(self, value):
     def _build_request(self):
         if self._next_url:
             parsed_next = urlparse(self._next_url)
+            # OData v2 allows relative __next values; resolve them against the service root.
+            if not parsed_next.scheme and not parsed_next.netloc:
+                resolved = urljoin(self._url, self._next_url)
+                parsed_next = urlparse(resolved)
```
filak-sap, line 303: I do not see the reason for not doing this:

```
url = urljoin(self._url, self._next_url)
```

the code below just checks the scheme and host and it comes from `self._url`.

pyodata/v2/service.py
```diff
@@ -297,13 +297,19 @@ def add_headers(self, value):
     def _build_request(self):
         if self._next_url:
             parsed_next = urlparse(self._next_url)
+            # OData v2 allows relative __next values; resolve them against the service root.
+            if not parsed_next.scheme and not parsed_next.netloc:
```
filak-sap, line 301: I would make the condition more strict. If the `_next_url` has scheme or netloc, then treat it as absolute URL and not just path.

filak-sap:
Please provide links to OData V2 specifications of the relative next link.

phanak-sap:
> Please provide links to OData V2 specifications of the relative next link.

My current understanding is this:
- such URLs would be possible to handle before the #308, now it will be rejected by the cross-origin check
- it is explicitly for Odata V4 spec - https://docs.oasis-open.org/odata/odata-json-format/v4.0/errata03/os/odata-json-format-v4.0-errata03-os-complete.html#_Toc453766625
- for odata v2 I am unable to find such exact wording, but in the end URI in _next fallback to same RFC 2616 - "URIs in HTTP can be represented in absolute form or relative to some  known base URI" (which could be according to me the empty service root). I also think this is not the area where odata v4 is backward incompatible with v2. 
 
If we agree on this being valid bugfix, I will address the other review comments.

filak-sap:
Please provide the evidence for your claim `OData v2 permits relative __next values.` or close this PR as 'not a bug'.

phanak-sap:
> Please provide the evidence for your claim `OData v2 permits relative __next values.` or close this PR as 'not a bug'.

OK. No problem here closing directly as for odata v2 it is vague claim.

PR #313, opened 2026-08-10
model: escape single quotes in Edm.String literals

lefranz:
Hi — thanks for maintaining pyodata.

### Problem

`EdmStringTypTraits.to_literal()` wraps the value in single quotes but does not escape quotes inside the value:

```python
>>> from pyodata.v2.model import EdmStringTypTraits
>>> EdmStringTypTraits().to_literal("délit d'initié")
"'délit d'initié'"          # the literal ends at the apostrophe
```

An embedded single quote in an OData string literal has to be doubled (`''`), so this produces a query the service refuses. Any keyword-argument or callable filter over a value containing an apostrophe is affected — which in practice means a lot of French, Italian and Irish text (`l'Etat`, `dell'`, `O'Brien`).

### Reproduction against a live service

Filtering the Swiss parliament's public OData v2 endpoint (`ws.parlament.ch`, via [swissparlpy](https://github.com/metaodi/swissparlpy)) on `Title eq "délit d'initié"`:

```
GET /odata.svc/Business?$filter=Language eq 'FR' and Title eq 'délit d'initié'
→ HTTP 400
  {"error":{"message":{"value":"Syntax error at position 45 in
   'Language eq 'FR' and Title eq 'délit d'initié''."}}}
```

Position 45 is the apostrophe in `d'initié`. With this patch the identical call returns **HTTP 200** and the expected row.

### Change

- `to_literal`: double any `'` in the value.
- `from_literal`: collapse `''` back to `'`, so the two directions round-trip. This felt in scope since #309 just tightened `from_literal`, but I am happy to drop it and keep this PR to the wire-format bug alone if you would rather treat the inbound direction separately.

### Notes on decisions

- `to_literal` now calls `str(value)` explicitly, because the `%s` interpolation that used to stringify implicitly is replaced by a `.replace()` call. Accepted types are unchanged.
- I did **not** touch `pyodata/v4/`, having only verified the v2 path.
- Escaping is applied in the traits rather than at the filter-construction sites, so it covers every caller that goes through `Edm.String` literals.

### Tests

Added to `test_traits` in `tests/test_model_v2.py`, next to the `from_literal` cases from #309: `to_literal` doubling, and a round-trip loop over `["O'Brien", "l'Etat", "a''b", "'wrapped'", 'no quotes', '']`.

Full suite passes: **295 passed**. `pylint`/`flake8` from `dev-requirements.txt` could not run in my environment (Python 3.12 — the pinned versions fail on the removed `pkg_resources` and on the `importlib.metadata` entry-points API); this is independent of the change, and no added line exceeds 120 columns.

CHANGELOG updated under `[Unreleased] → Fixed`.


filak-sap (approved):
Well done! Thank you very much for your contribution. Excellent summary helping us to save time re-learning the odata stuff.

